PR #241, opened 2015-03-09
Allow usage of sshpass

getreu:
Suggestion for enhancement for discussion:   Allow SSH password login

This class should then compose a command line like this: 
/usr/bin/sshpass -f ~/remotepasswordfile ssh -a -x -l root -oBatchMode=no -oConnectTimeout=15 -oBatchMode=yes -oStrictHostKeyChecking=no sysadm.local hostname

Part of the solution would be deleting this line:
``cmd_l.append("-oBatchMode=yes")``
(Is it really necessary?)

There is still another issue :
"""
clush: nodeset=sysadm.local fanout=64 [timeout conn=10.0 cmd=0.0] command="hostname"
SSH: /usr/bin/sshpass -f ~/remotepasswordfile ssh -a -x -l root -oBatchMode=no -oConnectTimeout=15 -oBatchMode=yes -oStrictHostKeyChecking=no sysadm.local hostname
clush: sysadm.local: exited with exit code 255
"""
The above SSH command line works in a shell but not invoked by clush.

### DIFF
--- a/lib/ClusterShell/Worker/Ssh.py
+++ b/lib/ClusterShell/Worker/Ssh.py
@@ -109,9 +109,6 @@ class ScpClient(CopyClient):
         if connect_timeout > 0:
             cmd_l.append("-oConnectTimeout=%d" % connect_timeout)
 
-        # Disable passphrase/password querying
-        cmd_l.append("-oBatchMode=yes")
-
         # Add custom scp options
         for opts in options:
             if opts:


thiell:
Hm. Another issue is that ssh_path is taken as a single command argument. A path.split(" ") is missing to fix that. So the right fix seems to be something like:
```python
cmd_l = path.split(" ")
# Add custom ssh options first as the first obtained value is used
       if options:
           cmd_l += options.split()
# Add other options there (overridable by user)
cmd_l += [ "-a", "-x"  ]
etc.
```

With these changes, I made sshpass work!

Note: Same kind of changes needed in `ScpClient`. Can you please provide a new patch?

getreu:
Thank you for investigating!
I think the clearcut solution would be to add to more options: ``sshpass-path`` and ``sshpass-options``. These will be prepended to ``ssh-path``. What do  you think?
Very important: ``-oBatchMode=yes`` must not stay hard coded! 
Btw.: sshpass-path, sshpass-options,  ssh-path, ssh-user and ssh-option should be settable via commandline also!?

PR #242, opened 2015-03-14
Support for sshpass authentification

getreu:
Non-interactivly performing password authentication with SSH's so called "interactive keyboard password authentication".

In order to use this feature the following options must be on command line or in
`/etc/clustershell/clush.conf`:

```
ssh_user: root
ssh_path: /usr/bin/sshpass -f /root/remotepasswordfile /usr/bin/ssh
scp_path: /usr/bin/sshpass -f /root/remotepasswordfile /usr/bin/scp
ssh_options: -oBatchMode=no -oStrictHostKeyChecking=no
```

In this example `/root/remotepasswordfile` is the file where the password belonging to the remote accounts for the ssh user `root` is stored.

### DIFF
--- a/conf/clush.conf
+++ b/conf/clush.conf
@@ -12,6 +12,16 @@ fd_max: 16384
 history_size: 100
 node_count: yes
 verbosity: 1
+
+# Add always all remote hosts to known_hosts without confirmation
 #ssh_user: root
 #ssh_path: /usr/bin/ssh
 #ssh_options: -oStrictHostKeyChecking=no
+
+# Non-interactivly performing password authentication with SSH's so called "interactive keyboard password authentication".
+#ssh_user: root
+#ssh_path: /usr/bin/sshpass -f /root/remotepasswordfile /usr/bin/ssh
+#scp_path: /usr/bin/sshpass -f /root/remotepasswordfile /usr/bin/scp
+#ssh_options: -oBatchMode=no -oStrictHostKeyChecking=no
+
+
--- a/lib/ClusterShell/Worker/Ssh.py
+++ b/lib/ClusterShell/Worker/Ssh.py
@@ -58,7 +58,12 @@ class SshClient(ExecClient):
         options = task.info("ssh_options")
 
         # Build ssh command
-        cmd_l = [ path, "-a", "-x"  ]
+        cmd_l = path.split()
+        # Add custom ssh options first as the first obtained value is used
+        if options:
+           cmd_l += options.split()
+        # Add other options there (overridable by user)
+        cmd_l += [ "-a", "-x"  ]
 
         if user:
             cmd_l.append("-l")
@@ -68,16 +73,10 @@ class SshClient(ExecClient):
         if connect_timeout > 0:
             cmd_l.append("-oConnectTimeout=%d" % connect_timeout)
 
-        # Disable passphrase/password querying
+        # Disable passphrase/password querying (overridable by user)
         cmd_l.append("-oBatchMode=yes")
-
-        # Add custom ssh options
-        if options:
-            cmd_l += options.split()
-
         cmd_l.append("%s" % self.key)
         cmd_l.append("%s" % self.command)
-
         return (cmd_l, None)
 
 class ScpClient(CopyClient):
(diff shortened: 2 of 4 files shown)


degremont:
Thanks for this patch which looks almost OK.
As you are modifying clush.conf options, you need to update the doc too.
Could you also fix the following files?

- vim syntax files (clush.conf.vim)
- txt man page (txt/clush.conf.txt), clush.conf.5 should be autogenerated with rst1man.py
- in a perfect world, the latex guide too
- tests: at least CLIConfigTest.py

getreu:
Sure the doc should be updated and the new feature should be mentioned somewhere.

I do not change the semantics or syntax of any existing option and I do not introduce any new. As this patch only obliterates an undocumented restriction of existing options it should do without additional documentation for the moment. It is just a matter of time, I am very busy at the moment. As soon as I have time I will help updating the documentation.

Concerning `CLIConfigTest.py`. I do not understand how it works. Can you help?

Related to this is request is https://github.com/cea-hpc/clustershell/pull/244. I am not happy with my solution and would like to improve it. Please give feed back.

degremont:
Actually you added 2 new options to clush.conf config file : ``scp_path`` and ``scp_options``. Even if they were supported internally, it was not possible to set them through those config files. In this regard, documentation needs to be updated.

In ``CLIConfigTest.py``, the simplest thing to do is to update ``testClushConfigFull()``. It writes a sample config file (``f.write(...)``), read it, and check the config values are correct. You will need to add more checks like ``self.assertEqual(config.scp_path, ...)``

``clushconf.vim`` is pretty straightforward to update.

man page needs a little bit more work. Guide is the hardest.

I understand this needs some time and you are busy, that's why I simply asked you add references to the 2 new option in doc files. No need to write a large explanation for them if you're too busy.

conf/clush.conf
```diff
@@ -12,6 +12,16 @@ fd_max: 16384
 history_size: 100
 node_count: yes
 verbosity: 1
+
+# Add always all remote hosts to known_hosts without confirmation
 #ssh_user: root
 #ssh_path: /usr/bin/ssh
 #ssh_options: -oStrictHostKeyChecking=no
+
+# Non-interactivly performing password authentication with SSH's so called "interactive keyboard password authentication".
```
thiell: - please split this comment into 2 lines
- typo in Non-interactiv**e**ly

lib/ClusterShell/Worker/Ssh.py
```diff
@@ -58,7 +58,12 @@ def _build_cmd(self):
         options = task.info("ssh_options")
 
         # Build ssh command
-        cmd_l = [ path, "-a", "-x"  ]
+        cmd_l = path.split()
```
thiell: Use shlex.split(path) - indeed shlex handles whitespaces and quotes, we should had used it before!

lib/ClusterShell/Worker/Ssh.py
```diff
@@ -58,7 +58,12 @@ def _build_cmd(self):
...
+        # Add custom ssh options first as the first obtained value is used
+        if options:
+           cmd_l += options.split()
```
thiell: use shlex.split(options)

lib/ClusterShell/Worker/Ssh.py
```diff
@@ -97,7 +96,13 @@ def _build_cmd(self):
         options = [ task.info("ssh_options"), task.info("scp_options") ]
```
thiell: use options = task.info("scp_options") or task.info("ssh_options")

lib/ClusterShell/Worker/Ssh.py
```diff
@@ -97,7 +96,13 @@ def _build_cmd(self):
...
 
         # Build scp command
-        cmd_l = [ path ]
+        cmd_l = path.split()
```
thiell: use shlex.split()

lib/ClusterShell/Worker/Ssh.py
```diff
@@ -97,7 +96,13 @@ def _build_cmd(self):
...
+
+        # Add custom ssh options first as the first obtained value is used
+        # Add custom scp options
+        for opts in options:
+            if opts:
```
thiell: same, use shlex.split()

thiell:
Thanks for your patch. We are almost done. The best option instead of split() is to use `shlex.split()` that does a better job. Also, I am ok with your suggestion of not using ssh_options for scp if scp_options is set. This can be written easily using `task.info("scp_options") or task.info("ssh_options")`.
Could you please provide an updated patch?

getreu:
Thank you for your detailed feedback. I'll update the patch next week-end.

degremont:
If you're too busy, skip the documentation part. We will do it later. 
It will be still very cool to accept the code you've improved!

thiell:
I'm preparing a patchset to add related (but basic) docs and to make some other improvements. Feel free to open a new pull request for other improvements, we would be happy to work with you again.

getreu:
Thank you, it was a pleasure :)

PR #244, opened 2015-03-15
Allow scp_options overwrite ssh_options

getreu:
scp uses ssh_options _and_ scp_options by default. In some usecases scp needs slightly different options then ssh. This is why scp_options was introduced. scp_options needs to be sourced first, in order to allow changes to ssh_options.

### DIFF
--- a/lib/ClusterShell/Worker/Ssh.py
+++ b/lib/ClusterShell/Worker/Ssh.py
@@ -94,7 +94,7 @@ class ScpClient(CopyClient):
         task = self.worker.task
         path = task.info("scp_path") or "scp"
         user = task.info("scp_user") or task.info("ssh_user")
-        options = [ task.info("ssh_options"), task.info("scp_options") ]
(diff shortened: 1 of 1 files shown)


getreu:
It is better then it was before but I am still not happy with the line
`options = [ task.info("scp_options"), task.info("ssh_options") ]`

It is 
* asymmetric
* hard to document
* hard to understand

Symmetric command line options should not interfere with each other.

Ideally it should be 
```
options = task.info("scp_options")
if options:
   cmd_l += options.split()
```

In order not to break existing config files I suggest the following instead:
```
ssh_options = task.info("ssh_options")
scp_options = task.info("scp_options")
if scp_options:
   options= scp_options
else:
   options=ssh_options
if options:
   cmd_l += options.split()
```

What do you think? If you agree I will write another pull request.

thiell:
Please see my comments in #242.

PR #248, opened 2015-04-24
Accept new options "ssh_path", "ssh_options", ... from commandline too....

### LINKED ISSUE #247: recently introduced new options --ssh_path, --scp_path, --ssh_options and --scp_options do not work from commandline
As part of my recent pullrequest #245 the above options had been modified respectively introduced. When I tested the options I defined them in the config file and everything worked well. 
Now I tried to submit them on the command line and they were not understood!

This possibility is important because users will use sshpass usualy once in the beginning of a session to append their certificates to the remote authorized_hosts file. Once this is done sshpass is not needed any more as clush can connect the usual way (which is much faster). In order to not be forced to modify the configuration file just for this one and first connect, users should be enabled to pass   --ssh_path, --scp_path, --ssh_options and --scp_options on the command line.

My pull request "support for ssh_pass added #245" adds some code for this but it seems it is not complete. I do not have time now, maybe someone could have a look what is missing. I am sure it is just a small thing.

getreu:
... Closes #247.

.Test
`> clush --ssh_path='sshpass -p <MySecretPassword> ssh' --ssh_options='-oBatchMode=no -oStrictHostKeyChecking=no' -l <MyRemoteUser> -w <OtherHost> echo hallo`

This possibility is important because users will use sshpass usualy once in the beginning of a session to append their certificates to the remote authorized_hosts file. Once this is done sshpass is not needed any more as clush can connect the usual way (which is much faster). In order to not be forced to modify the configuration file just for this one and first connect, users should be enabled to pass --ssh_path, --scp_path, --ssh_options and --scp_options on the command line.

.Use case: Distribute and install ssh-keys
`> cat ~/.ssh/id_rsa.pub | clush --ssh_path='sshpass -p <MySecretPassword> ssh' --ssh_options='-oBatchMode=no -oStrictHostKeyChecking=no' -l <MyRemoteUser> -w <OtherHosts> -b 'cat - >>/home/<MyRemoteUser>/.ssh/authorized_keys'`

### DIFF
--- a/lib/ClusterShell/CLI/Config.py
+++ b/lib/ClusterShell/CLI/Config.py
@@ -92,8 +92,14 @@ class ClushConfig(ConfigParser.ConfigParser, object):
             self._set_main("fanout", options.fanout)
         if options.user:
             self._set_main("ssh_user", options.user)
-        if options.options:
-            self._set_main("ssh_options", options.options)
+        if options.ssh_options:
+            self._set_main("ssh_options", options.ssh_options)
+        if options.ssh_path:
+            self._set_main("ssh_path", options.ssh_path)
+        if options.scp_options:
+            self._set_main("scp_options", options.scp_options)
+        if options.scp_path:
+            self._set_main("scp_path", options.scp_path)
         if options.connect_timeout:
             self._set_main("connect_timeout", options.connect_timeout)
         if options.command_timeout:
--- a/lib/ClusterShell/CLI/OptionParser.py
+++ b/lib/ClusterShell/CLI/OptionParser.py
@@ -202,8 +202,20 @@ class OptionParser(optparse.OptionParser):
                           help=optparse.SUPPRESS_HELP, type="float")
         optgrp.add_option("-l", "--user", action="store", type="safestring",
                           dest="user", help="execute remote command as user")
-        optgrp.add_option("-o", "--options", action="store", dest="options",
-                          help="can be used to give ssh options")
+        optgrp.add_option("-o", "--options", action="store", dest="ssh_options",
+                          help="set ssh options")
+        optgrp.add_option("--ssh_options", action="store",
+                          dest="ssh_options", help="set ssh " \
+                          "options, same as --options")
+        optgrp.add_option("--ssh_path", action="store",
+                          dest="ssh_path", help="set path to ssh, " \
+                          "can be used to prepend sshpass")
+        optgrp.add_option("--scp_options", action="store",
+                          dest="scp_options", help="set scp options", \
+                          type="float")
+        optgrp.add_option("--scp_path", action="store",
+                          dest="scp_path", help="set path to scp, " \
+                          "can be used to prepend sshpass")
         optgrp.add_option("-t", "--connect_timeout", action="store",
                           dest="connect_timeout", help="limit time to " \
                           "connect to a node" ,type="float")


getreu:
Please see my blog about clustershell here: http://getreu.net/public/downloads/doc/Secure_Computer_Cluster_Administration_with_SSH/#distribute-keys-clustershell

getreu:
This patch is pending for almost a month now. Could you please give some feed-back? Thank you.

degremont:
We do not want to add too many options to clush, especially adding a new option for its possible values in clush.conf. This should be managed differently.

We should probably add a generic way to change any config option on command line.
Either something like
~~~
 clush --options='ssh_path=value' ...
~~~
or a generic catch all for any unknown option. In this case, they will be tested against the config options
~~~
clush --ssh_path=value ...
~~~

Your final use case is a bit surprising to me. You wanted to have an automatic way to add a ssh key, password-less, to use it only once. This looks overkill...

getreu:
Bonjour,

thanks for your feedback.

On 21/05/15 12:23, Aurélien Degrémont wrote:
> We do not want to add too many options to clush, especially adding a
> new option for its possible values in clush.conf. This should be
> managed differently.
> 
> We should probably add a generic way to change any config option on 
> command line. Either something like
> 
> | clush --options='ssh_path=value' ... |
> 
> or a generic catch all for any unknown option. In this case, they
> will be tested against the config options
> 
> |clush --ssh_path=value ... |

This is what my patch does, isn't it so?

> Your final use case is a bit surprising to me. You wanted to have an 
> automatic way to add a ssh key, password-less, to use it only once.
> This looks overkill...

This is the way I am currently working with parallel-ssh: First I
install my ssh-key in the cluster with some special options using
sshpass. For the actual work I authenticate without sshpass. This is
_much_ faster and more secure.

The parallel-ssh workflow is described here:

http://getreu.net/public/downloads/doc/Secure_Computer_Cluster_Administration_with_SSH/#distribute-keys


The suggested clustershell workflow is described here:

http://getreu.net/public/downloads/doc/Secure_Computer_Cluster_Administration_with_SSH/#distribute-keys-clustershell

Without this patch will I have to temporarily replace clustershell's
config file for just one command.

Best regards



Jens Getreu

degremont:
>> |clush --ssh_path=value ... |
> This is what my patch does, isn't it so?

My proposal is to deal with those args, dynamically, not explicitly declaring each of them in the code. Imagine if we want to support 10 more new options, we want to avoid doing the same patch of all of them.

> For the actual work I authenticate without sshpass. This is _much_ faster and more secure.

This is also the same of us. We deploy a ssh key to before using with clustershell. And we never add such needs like sshpass for that. That key is either deployed by hand is there is few nodes or through our install/deploy system which bootstrap the nodes and deploy the correct ssh key.

> The suggested clustershell workflow is described here

I do not understand why you create key pairs on remote node in step #3. Did you mean to do see locally instead?

getreu:
>> My proposal is to deal with those args, dynamically, not explicitly declaring each of them in the code. Imagine if we want to support 10 more new options, we want to avoid doing the same patch of all of them.

For security reasons I strongly recommend to define them explicitly. It is only 2 lines per command line option.

getreu:
Any news here?

thiell:
Hi Jens,

What @degremont means is that it would be best to add only one option to clush that allows the user to override any config key/value pair defined in /etc/clustershell/clush.conf. And I fully agree, it is important than "clush -h" remains as light/easy/comprehensive as possible. However, I fully understand what you're asking and I thank you for making use of clustershell in your blog examples. We will make sure to find a solution with clush 1.7 so you won't need your patched version anymore. :)

@degremont: I don't like much the generic catch-all option, but I will be ok to add a `--config-override='ssh_path=value'` or `-O 'ssh_path=value'` that works with any present/future clush.conf options (ssh_path, scp_path, fd_max etc.). Some will be redundant like fanout, connect_timeout, command_timeout, user, but it's not a big deal.

Jens, if you would like to submit a new patch, I don't know if you've already played with gerrit but we have now a preference for working on GerritHub ( https://review.gerrithub.io/#/q/project:cea-hpc/clustershell ). To submit a patch for review, you can push a patch on the refs/for/master branch of the following git url:

   ssh://yourgerrithublogin@review.gerrithub.io:29418/cea-hpc/clustershell

You just need to install first the Change-Id git hook to be able to push there (help here http://stackoverflow.com/questions/8845658/gerrit-error-when-change-id-in-commit-messages-are-missing).

getreu:
Hello,

thank for your news. 

I wonder if the suggested syntax `-O 'ssh_path=value'` needs additional ticks. 
Would it still be possible to invoke _clush_ with the following parameters?

```bash
cat /root/.ssh/id_rsa.pub | clush --ssh_path='sshpass -p <OtherHostsPassword> ssh' --ssh_options='-oBatchMode=no -oStrictHostKeyChecking=no' -l root -w <OtherHosts> -b 'cat - >>/root.ssh/authorized_keys; chmod 600 /root.ssh/authorized_keys'
```

Unfortunately I do not have time to submit new patches until end of this year. Could you please leave note here when your above solution is ready? I will update my blog then.

getreu:
Hello! Any news here?

thiell:
Hi Jens,

I just finished the patch for this feature. It will now enter the code review phase (https://review.gerrithub.io/250653) and hopefully should be landed for 1.7.

I tested your command (slighty modified) and it works well.
With the current patch, the syntax will be as follow:

```console
$ cat /root/.ssh/id_rsa.pub | clush -O ssh_path='sshpass -p <OtherHostsPassword> ssh' \
                                    -O ssh_options='-oBatchMode=no -oStrictHostKeyChecking=no' \
                                    -l root -w <OtherHosts> -b 'cat - >>/root.ssh/authorized_keys; chmod 600 /root.ssh/authorized_keys'
```

getreu:
Hello!

Thanks for implementing this feature! I had no time to test it yet, but I already updated my 
[Cluster admin howto](http://getreu.net/public/downloads/doc/Secure_Computer_Cluster_Administration_with_SSH/#distribute-keys-clustershell).

PR #252, opened 2015-05-26
Task: add default event handler in task.shell()

brianpkelly:
Fixed the bug I ran into as detailed in this [thread](http://sourceforge.net/p/clustershell/mailman/clustershell-devel/thread/CAMD27Oi836s4Y8vQn6A3sSzrFipzasHhQ-hOw28hSJ26knCAgA%40mail.gmail.com/#msg34136205) from the mailing list.

### DIFF
--- a/lib/ClusterShell/Task.py
+++ b/lib/ClusterShell/Task.py
@@ -577,7 +577,7 @@ class Task(object):
         >>> task.resume()
         """
 
-        handler = kwargs.get("handler", None)
+        handler = kwargs.get("handler", EventHandler())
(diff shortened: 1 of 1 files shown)


degremont:
Thanks for this pull request but this bug should not be fixed that way.
It is perfectly normal for a worker to not have an event handler.

The bug only appears when using the new gateway code. The proper fix is more removing event handler checks in `PropagationChannel.recv_ctl()` but I need to look at this more closely to be sure...

brianpkelly:
Ah OK, I wasn't sure whether an event handler was necessary or not.

PR #273, opened 2015-10-28
Prevent padding groups errors

mdavranche:
Grouping similar nodes with different 0 padding leads to error.

Before this patch:
server1 server2 server03 are grouped to server\[01-03\] (what is wrong)

After this patch:
server1 server2 server03 are grouped to server[1-2],server03

### DIFF
--- a/lib/ClusterShell/NodeSet.py
+++ b/lib/ClusterShell/NodeSet.py
@@ -778,7 +778,7 @@ class ParsingEngine(object):
         Initialize Parsing Engine.
         """
         self.group_resolver = group_resolver
-        self.base_node_re = re.compile("(\D*)(\d*)")
+        self.base_node_re = re.compile("([^1-9]*)(\d*)")
 
     def parse(self, nsobj, autostep):
         """


thiell:
Hi, thank for your pull request.

I tried this patch but it breaks folding and other features:

For example, before this patch:
```
$ nodeset.py -f server03 server11
server[03,11]
```

With this patch:
```console
$ nodeset.py -f server03 server10
server10,server03
```

BTW we have regression tests available in tests and you can see that it breaks too many things:
```console
$ cd tests
$ nosetests -sv NodeSetTest.py
```

We're already aware of the problem. Different 0 padding for a same "pattern" or cluster name (here: server%s) is difficult to handle right now as it is managed by RangeSet. And we still need all set operations to work (eg. server[06-13]&server[10-16] should return server[10-13], etc).

However, I fully agree that your example makes sense and should probably be fixed. We'll use your pull request ticket to track such improvement. Just a question, is it a real use case for you to mix server2 and server03 for instance? ( Isn't that error prone? )

mdavranche:
Indeed, this patch breaks grouping optimisations...
Without this patch: most of grouping are true and optimized, but few are wrong
With this patch: all the grouping are true, but lots are not optimized
It depends on what we want ;)

Like you said, solving this is not easy, I can't find easy way to do, so I understand this pull request is not suitable for merging, this is why I close it. Be free to re-open it. Note that I will personnaly use the patch for my case.

To answer your question, yes, this is a real case. We created server1 and server2 because we didn't expect the platform to grow that way. We will rename them server01 and server02, but waiting for this, we have to deal with those names.

thiell:
Thank you for the feedback, that makes sense!

Issue #239, opened 2015-02-27
Drop support of Python 2.4 and 2.5 to prepare Python 3


degremont:
Python 3 will be default Python version for Fedora 22. This does not means Python 2 tools won't work any more but it will become more difficult to support them.

We should start working toward supporting Python 3. The first step for that is to stop supporting Python 2.4 and 2.5. 
In RHEL world, Python 2.6 is there since RHEL 6.0. Fedora supports Python 2.6 since ages. Debian stable uses Python 2.7 and was previously already supporting Python 2.6. Same for Ubuntu. Arch, Gentoo are fine too.

We should removing all deprecated stuff from these versions to come closer to Python 3 syntax.
Here is short list of things to start with:
- except syntax (`except Exception as exp`)
- all classes must inherit from `object`

`pylint --disable=all --enable=old-style-class  ClusterShell/`
- drop `has_key()`
- relative import should be fine (I don't think we are using them. Pylint normally complains if we do so)
- several methods now return an iterator instead of a list. This could easily trigger bugs, but hard to catch them.
- unittest compability (see 2to3)

An interesting documentation: https://docs.python.org/3.3/howto/pyporting.html
https://docs.python.org/3/library/2to3.html#2to3fixer-except
http://python-modernize.readthedocs.org/en/latest/fixers.html

Strings management will be a pain. I think this should be done at the end.
Prefer making small patches to do this incrementally and not a big patch for everything.

thiell:
Another thing to look at is the use of `super()`.

As for Python 3, not that I do not work on it... :), but it's not yet the default on Fedora 22 (I have 2.7.10 right now on a fresh install).

thiell:
We could use [bytes](https://www.python.org/dev/peps/pep-0358/) instead of str whenever possible. It is available in Python 2.6 as an alias of str.

degremont:
Using bytes should probably be done when fixing #22 and #309

SyntaxRules:
+1 For python 3 support

degremont:
This ticket looks very close to be finished. Just need to update README & docs to announce we only support from Python 2.6 now, and we can close it.

PR #298, opened 2016-03-25
Update exception syntax.

wt:


lib/ClusterShell/NodeSet.py
```diff
@@ -269,7 +269,7 @@ def _iter_nd_pat(self, pat, rset):
                 # set of user-provided fold axis (support negative numbers)
                 fold_axis = [int(x) % dimcnt for x in self.fold_axis
                              if -dimcnt <= int(x) < dimcnt]
-        except (TypeError, ValueError), exc:
+        except (TypeError, ValueError) as exc:
```
thiell: Hey @wt thanks for this, appreciated! Now I know this can be frustrating, but `except ... as` only works from Python 2.6 and CS 1.7 still supports Python 2.4 - the current syntax works from Python 2.4 to 2.7. We plan to drop Python 2.4 (so also to drop EPEL 5 support) in ClusterShell 1.8+ and work on Python 3 support. It may be time to create a branch, as we have still a few bugfix patches to land in a v1.7.2...

wt:
Good to hear that 1.8 will port to python3. Will you be using the "six" lib to maintain backward compat?

degremont:
1.8 will not support Python 3 yet. 1.8 is dropping support of Python 2.4 and 2.5, first step before going to a real support for Python 3, probably after 1.8 :)
Actually, issue #239 will be implemented in 1.8, and your pull request is a good fit for this ticket.

wt:
Interesting, maybe there is a way to maintain python2.4 compat with the exception syntax if you have any interest in that:
http://stackoverflow.com/questions/12682558/how-to-write-an-exception-catching-code-works-in-python2-4-to-python3

degremont:
Thanks for this idea. 
Target is really to ease support for Python 3+ in the future. Exception is not the only thing that need to be handled. We do not really care about supporting 2.4 anymore after v1.7
So its is fine to simply moved to the new syntax when 1.8 development will be started.

thiell:
Closing as now very deprecated. 1.8 will fully support Python 2.6+ and Python 3.4+
Please check 1.8 beta 1 (1.7.81) available on pypi and epel-testing and open PR or Issues if needed.
Thanks!

PR #330, opened 2016-12-08
Fix documentation for NodeSet difference_update()

volans-:
The difference_update() methods in NodeSet and NodeSetBase classes don't return any object, they update the current instance. Modified the documentation accordingly.

### DIFF
--- a/lib/ClusterShell/NodeSet.py
+++ b/lib/ClusterShell/NodeSet.py
@@ -634,8 +634,7 @@ class NodeSetBase(object):
 
     def difference_update(self, other, strict=False):
         """
-        ``s.difference_update(t)`` returns nodeset s after removing
-        elements found in t.
+        ``s.difference_update(t)`` remove from s all the elements found in t.
 
         :raises KeyError: an element cannot be removed (only if strict is
             True)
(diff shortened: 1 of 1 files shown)


lib/ClusterShell/NodeSet.py
```diff
@@ -634,8 +634,7 @@ def __sub__(self, other):
 
     def difference_update(self, other, strict=False):
         """
-        ``s.difference_update(t)`` returns nodeset s after removing
-        elements found in t.
+        ``s.difference_update(t)`` remove from s all the elements found in t.
```
degremont: just wondering if this should not be written "remove**s** from s"

lib/ClusterShell/NodeSet.py
```diff
@@ -1484,9 +1483,9 @@ def intersection_update(self, other):
 
     def difference_update(self, other, strict=False):
         """
-        s.difference_update(t) returns nodeset s after removing
-        elements found in t. If strict is True, raise KeyError
-        if an element cannot be removed.
+        s.difference_update(t) remove from s all the elements
```
degremont: same here

lib/ClusterShell/NodeSet.py
martinetd: Either `s.difference_update(t) removes`... or ^`remove from s` without subject (infinitive)

thiell:
Hi @volans-, could you update your PR so we can include this patch in 1.7.3?

volans-:
@thiell Done.

thiell:
@volans- thanks!

PR #341, opened 2017-05-16
Ssh: use long-format hardwired ssh options (#340) 

### LINKED ISSUE #340: Hardwired options are not overridable on commandline.
In this commit
https://github.com/cea-hpc/clustershell/blob/6e4e131f454b2ac37cfcd93e1a17d9a39bbbe951/lib/ClusterShell/Worker/Ssh.py#L60
there is a comment
> Add custom ssh options first as the first obtained value is used. Thus all options are overidable by custom options. 

Looks like it is not the case:
```
[alex@host ClusterShell]$ ssh -A -a  remotehost 'env | grep SSH_AUTH_SOCK; echo $?' 
1
[alex@host ClusterShell]$ ssh -a -A  remotehost 'env | grep SSH_AUTH_SOCK; echo $?' 
SSH_AUTH_SOCK=/tmp/ssh-LsWLOAXZGe/agent.10338
0
```
When we specify `-a`(not forward agent), and then `-A`(forward agent) it is forwarded.
You can also play with order of `-q` and `-v` to confirm. Last specified option wins, because they iterate over options in a while loop https://github.com/openssh/openssh-portable/blob/V_7_5_P1/ssh.c#L608

Hence this doesn't work as expected:
```
clush -d -o "-A" -w  remotehost 'env | grep SSH_AUTH_SOCK; echo $?'
clush: nodeset=remotehost fanout=64 [timeout conn=15.0 cmd=0.0] command="env | grep SSH_AUTH_SOCK; echo $?"
SSHCLIENT: 
ssh -A -a -x -oConnectTimeout=15 -oBatchMode=yes remotehost env | grep SSH_AUTH_SOCK; echo $?
...
1
```

I think this statement about order came from here:

```
SSH_CONFIG(5)                                                          BSD File Formats Manual                                                         SSH_CONFIG(5)

NAME
     ssh_config — OpenSSH SSH client configuration files

SYNOPSIS
     ~/.ssh/config
     /etc/ssh/ssh_config

DESCRIPTION
     ssh(1) obtains configuration data from the following sources in the following order:

           1.   command-line options
           2.   user's configuration file (~/.ssh/config)
           3.   system-wide configuration file (/etc/ssh/ssh_config)

     For each parameter, the first obtained value will be used. 
```

Looks like that means that value in 1. will override value in 2, and value in  2. overrides 3.
That also true for order of options in the config file.
E.g, this result to forwarding enabled:
```
Host *
  ForwardAgent yes
  ForwardAgent no
```
This result to forwarding disabled:
```
Host *
  ForwardAgent no
  ForwardAgent yes
```

But since it is not the case for cli, I think adding custom options should be done in the very end, to let user effectively override everything hardwired, unless you think it should be fixed in openssh-client.

I used these versions:
```
[root@scrmca01arbco001 alex]# rpm -qa  | grep -e openssh-c -e clustershell
openssh-clients-6.6.1p1-31.el7.x86_64
clustershell-1.7.2-1.el7.noarch
```

alexbozhenko:
man ssh-config says that "For each parameter, the first obtained value will
be used", but it turns out that it is not the case for short format ssh
options, like '-x' or '-a'. For them last specified value wins, making
impossible for user to override those. Using only long format options will
let users override hardwired values.

Closes #340

### DIFF
--- a/lib/ClusterShell/Worker/Ssh.py
+++ b/lib/ClusterShell/Worker/Ssh.py
@@ -57,7 +57,8 @@ class SshClient(ExecClient):
             cmd_l += [os.path.expanduser(opt) for opt in shlex.split(options)]
 
         # Hardwired options (overridable by ssh_options)
-        cmd_l += [  "-a", "-x"  ]
+        # note: you should use only long-format options here
+        cmd_l += ["-oForwardAgent=no", "-oForwardX11=no"]
 
         if user:
             cmd_l.append("-l")
@@ -105,7 +106,7 @@ class ScpClient(CopyClient):
             # use expanduser() for options like '-i ~/.ssh/my_id_rsa'
             cmd_l += [os.path.expanduser(opt) for opt in shlex.split(options)]
 
-        # Hardwired options (overridable by ssh_options)
+        # Hardwired options
         if self.isdir:
             cmd_l.append("-r")
 


lib/ClusterShell/Worker/Ssh.py
```diff
@@ -57,7 +57,8 @@ def _build_cmd(self):
             cmd_l += [os.path.expanduser(opt) for opt in shlex.split(options)]
 
         # Hardwired options (overridable by ssh_options)
-        cmd_l += [  "-a", "-x"  ]
+        # note: you should use only long-fromat options here
```
degremont: (typo) "format"

lib/ClusterShell/Worker/Ssh.py
```diff
@@ -57,7 +57,8 @@ def _build_cmd(self):
...
+        cmd_l += [  "-oForwardAgent=no", "-oForwardX11=no"  ]
```
degremont: (style) could you remove the useless whitespace (PEP8) around brackets : ["-oFor...", "-oFor..."]

alexbozhenko:
@degremont Thanks for the review. I updated my commit.

PR #354, opened 2017-09-03
EnginePoll/EngineEpoll: fix polling timeout

### LINKED ISSUE #353: Python 3 on macOS/BSDs: Clush and library fails with no timeout
I wanted to test the latest release to start migrating my tool to Python 3 (thanks a lot for adding Py3 support!) and found that both `clush` and `ClusterShell` as a library fails on `macOS` (and other BSDs) when not setting a timeout, with `ERROR: [Errno 22] Invalid argument`.

After digging a bit I discovered that:
- in Python 3 the `Engine.Factory.PreferredEngine` factory returns an `EnginePoll` on macOS, while on Python 2 it was returning an `EngineSelect` because `poll()` was not supported.
- when polling for events in `Engine.Poll.EnginePoll.runloop()`, the timeout passed to the `poll.poll()` method is `timeo * 1000.0 + 1.0`, and when the timeout is not specified, `timeo = -1`, hence the passed timeout is `-999.0`.
- Python passes the timeout parameter of `select.poll.poll([timeout])` basically as-is down to the `poll()` system call
- On macOS and other BSDs the `poll()` system call throws an error on arbitrary negative values. Only non-negative integers and `-1` (or the value of `INFTIM` if defined) are accepted.

I've already opened an issue to Python [1] and sent a PR to hopefully fix it [2], but I'll follow up with a PR to ClusterShell too to workaround this issue passing directly `-1` to the `poll()` method.

[1] https://bugs.python.org/issue31334
[2] python/cpython/pull/3277

volans-:
Force the timeout to be exactly `-1` when is negative. The implementation
of the `poll()` system call in some OSes, including macOS and BSDs, throws
an error for arbitrary negative values different from `-1`.

Fix #353

### DIFF
--- a/lib/ClusterShell/Engine/Poll.py
+++ b/lib/ClusterShell/Engine/Poll.py
@@ -109,7 +109,12 @@ class EnginePoll(Engine):
                     timeo = timeout
 
                 self._current_loopcnt += 1
-                evlist = self.polling.poll(timeo * 1000.0 + 1.0)
+
+                if timeo < 0:
+                    poll_timeo = -1
+                else:
+                    poll_timeo = timeo * 1000.0
+                evlist = self.polling.poll(poll_timeo)
 
             except select.error as ex:
                 # might get interrupted by a signal
--- a/lib/ClusterShell/Engine/EPoll.py
+++ b/lib/ClusterShell/Engine/EPoll.py
@@ -114,7 +114,12 @@ class EngineEPoll(Engine):
                     timeo = timeout
 
                 self._current_loopcnt += 1
-                evlist = self.epolling.poll(timeo + 0.001)
+
+                if timeo < 0:
+                    poll_timeo = -1
+                else:
+                    poll_timeo = timeo
+                evlist = self.epolling.poll(poll_timeo)
 
             except IOError as ex:
                 # might get interrupted by a signal


lib/ClusterShell/Engine/Poll.py
```diff
@@ -109,7 +109,12 @@ def runloop(self, timeout):
                     timeo = timeout
 
                 self._current_loopcnt += 1
-                evlist = self.polling.poll(timeo * 1000.0 + 1.0)
+
+                if timeo < 0:
+                    poll_timeo = -1
+                else:
+                    poll_timeo = timeo * 1000.0 + 1.0
```
degremont: I think the only code path to get a buggy value for this math is when timeo == -1 above (line 108).
The only place where this math should be done is when timeo is positive, so on line 107. 

I think you should simply move this math on line 107 and simply pass timeo to `poll()`

By the way, EPoll should have the same logic. I do not know if it has the same bug, but it is a good idea to also fix it the same way.

volans-: @degremont thanks for the quick review. The reason why I did it there is that `timeo = timeout` is also set on line 109 if `timeo == -1` and `timeout` is a parameter of the `runloop` method, that could potentially be negative. I didn't want to make assumptions related to the result of the `nextfire_delay()` call and the possibility to have a negative timeout passed, so erring on the safe side.
If you tell me that is safe to move the math logic to line 107 I'll do it.

Regarding `EPoll`, that's a Linux-only system call and from a quick test although the man page for `epoll_wait` states that `-1` should be specified, seems that it accept any negative value like the `poll` system call. Anyway is indeed safer to pass exactly `-1` to wait indefinitely. I'll include this change in the PR.

degremont: @volans- I've checked values returned by `nextfire_delay()` and possible values for `timeout`. 
Actually my proposal was wrong if `nextfire_delay()` returns -1 and `timeout` is positive.
After testing various possibilities of improvements, I could not find a nice way to write this.

I could not get something better than you. Please just update EPoll too.

@thiell You have a better knowledge than me of this part of the code. Maybe you will have a suggestion?

volans-: @degremont ok, PR updated.
I was also wondering if the case of adding 1 millisecond was to avoid to pass `0` as timeout to the `poll()` methods if/when `nextfire_delay()` returns `0`, given that `0` has a special meaning in the `poll()` methods. As of the current implementation `0` will never be passed as `timeout`, just checking that is the intended behaviour given that we're at it.

thiell: @volans- please get rid of the added millisecond, I don't recall any good reason to keep it. Then `0` will be passed to `poll()` or `epoll_wait()` when `self.timerq.nextfire_delay()` returns `0`, which may happen in case of laggy timers, but that's fine, the call will be non-blocking and should trigger any pending events + process pending timers.

The patch looks good to me, and yes, please update EPoll.py too. Thanks much!

volans-: @degremont @thiell done.

PR #362, opened 2017-10-11
Update Slurm bindings

kcgthb:
* shortcuts for group names
  * better filtering of states and partition names with '*'
  * added job bindings

### DIFF
--- a/conf/groups.conf.d/slurm.conf.example
+++ b/conf/groups.conf.d/slurm.conf.example
@@ -18,15 +18,15 @@ reverse: sinfo -h -N -o "%R" -n $NODE
 [slurmstate,st]
 map: sinfo -h -o "%N" -t $GROUP
 all: sinfo -h -o "%N"
-list: sinfo -h -o "%T" | sed 's/[*$]//g'
-reverse: sinfo -h -N -o "%T" -n $NODE | sed 's/[*$]//g'
+list: sinfo -h -o "%T" | tr -d '*~#$@+'
+reverse: sinfo -h -N -o "%T" -n $NODE | tr -d '*~#$@+'
 cache_time: 60
 
 #
 # SLURM job bindings
 #
 [slurmjob,sj]
-map: squeue -j $GROUP -h -o "%N"
+map: squeue -h -j $GROUP -o "%N"
 list: squeue -h -o "%i" -t R
-reverse: squeue -w $NODE -h -o "%i"
+reverse: squeue -h -w $NODE -o "%i"
 cache_time: 60
--- a/conf/groups.conf.d/slurm.conf.example
+++ b/conf/groups.conf.d/slurm.conf.example
@@ -3,19 +3,30 @@
 # Please see `man 5 groups.conf` for further details.
 #
 
+#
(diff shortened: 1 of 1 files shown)


conf/groups.conf.d/slurm.conf.example
```diff
@@ -3,19 +3,30 @@
...
 #
-[slurmstate]
-map: if [ "$GROUP" == "broken" ]; then sinfo -h -o "%N" -t down,drained; else sinfo -h -o "%N" -t $GROUP; fi
+[slurmstate,st]
+map: sinfo -h -o "%N" -t $GROUP
 all: sinfo -h -o "%N"
-list: sinfo -h -o "%T" | tr -d '~#$*'
-reverse: sinfo -h -N -o "%T" -n $NODE
-cache_time: 300
+list: sinfo -h -o "%T" | sed 's/[*$]//g'
```
degremont: Could you explain why this sed is better? Do you have examples?

conf/groups.conf.d/slurm.conf.example
```diff
@@ -3,19 +3,30 @@
...
+reverse: sinfo -h -N -o "%T" -n $NODE | sed 's/[*$]//g'
+cache_time: 60
+
+#
+# SLURM job bindings
+#
+[slurmjob,sj]
+map: squeue -j $GROUP -h -o "%N"
```
degremont: if you refresh this patch, keep `-h` at the beginning of each command. It makes things simpler to understand.

conf/groups.conf.d/slurm.conf.example
kcgthb: Although the current regex doesn't use it, `sed` could allow to only remove the last character of the string if it matches the blacklist, while `tr` will blindly remove all occurrences. Granted, those characters don't really happen in node state names, but what if Slurm someday introduces a "SuperB0o$t" state, huh? :grin:

Performance wise, `sed` is anecdotally faster than `tr`, but the `tr` syntax is maybe easier to read.
So, for readability's sake, let's stick with `tr`. And add `@` and `+` to the list of characters to remove. From the `sinfo` man page:
```
NODE STATE CODES
       Node state codes are shortened as required for the field size.  These node states may be followed by a special character to identify state flags associated with the node.  The following node sufficies and states are used:
       *   The node is presently not responding and will not be allocated any new work.  If the node remains non-responsive, it will be placed in the DOWN state (except in the case of COMPLETING, DRAINED, DRAINING, FAIL, FAILING nodes).
       ~   The node is presently in a power saving mode (typically running at reduced frequency).
       #   The node is presently being powered up or configured.
       $   The node is currently in a reservation with a flag value of "maintenance".
       @   The node is pending reboot.
       ALLOCATED   The node has been allocated to one or more jobs.
       ALLOCATED+  The node is allocated to one or more active jobs plus one or more jobs are in the process of COMPLETING.
       COMPLETING [...]
```

conf/groups.conf.d/slurm.conf.example
kcgthb: Good point.

Issue #372, opened 2018-01-30
ClusterShell/Task.py - _pchannel method calls logger.info could be logger.debug


bamb0u:
Hello guys,

we have a code which uses logger.info for default/normal logging. When we execute it, we inherit your messages  in _pchannel methodes & _pchannel_release, about creating and destroying channels. I'm wondering if it is really significant for you to log them using .info and if not, if you could instead use .debug  ?

Regards, Benjamin.

degremont:
We could probably change these message to info, but I think you can easily remove them changing the default log level for `ClusterShell.Task` logger, don't you?

bamb0u:
As far as I understand, our goal is to set the same level for our code and ClusterShell. If we ran our code in debug we will set the level of ClusterShell in debug.
By default we run with level info so we have your messages in these methods which seems the only case where you run logger.info.

volans-:
If this is the case you could also just set ClusterShell log level to debug when you use debug in your software and to warning when you use info in your software.

bamb0u:
It could be a workaround yes, but for now it simple to just use the same variable to set it.
And it seems there is no point to keep these 2 messages in info level in clustershell, don't you think ?
Let's think for the long run ;-)

degremont:
As I said, I agree with that, was just a quick workaround for you, at short term :)

PR #373, opened 2018-02-05
Task: pchannel logging updated from info to debug

bamb0u:
issue #372


bamb0u:
@degremont Ok, I have done it... :snowflake:

PR #375, opened 2018-02-06
Add missing argument of ev_hup in 1.8 release note

ThomasBoni:


degremont:
Thanks for the fix.
Could you update your patch to follow the same conventions for commit message than similar ones in Git history, like 4845532 or c3bd662?

ThomasBoni:
Updated in #376 with a well formatted commit message

PR #403, opened 2019-04-07
doc/sphinx: disable smartquotes

### LINKED ISSUE #402: doc: '--' is translated into '–' on html documentation
This issue was initially [found](https://github.com/cea-hpc/modules/issues/250) on [Modules](https://github.com/cea-hpc/modules) online documentation. But as you also use Sphinx I think it could also be interesting for clustershell.

By default Sphinx applies a `smartquotes` (also called `smartypants`) mechanism to convert quotes and dashes to lang-specific entities. This is kinda bad for command-line description as it automatically translates `--` into `–`, which most of the times leads to an error on tools.

It seems that it does not affect clustershell documentation much. After a quick look, I only found [one issue in the release note document](https://clustershell.readthedocs.io/en/latest/release.html#nodeset-l-list-all-option), where the `--list-all option` is written `–list-all option`.

xdelaruelle:
Disable `smartquotes` mechanism (called `html_use_smartypants` in
older Sphinx versions) to prevent quotes and dashes to be converted to
lang-specific entities.

Closes #402.

### DIFF
--- a/doc/sphinx/conf.py
+++ b/doc/sphinx/conf.py
@@ -140,7 +140,14 @@ def setup(app):
 
 # If true, SmartyPants will be used to convert quotes and dashes to
 # typographically correct entities.
-#html_use_smartypants = True
+# Better to disable this to prevent alteration of command-line arguments
+# Sphinx <1.6 use `html_use_smartypants` and >=1.6 use `smartquotes`
+from sphinx import __version__ as sphinx_version
+sphinx_version_parts = [int(i) for i in sphinx_version.split('.')]
+if sphinx_version_parts[0] <= 1 and sphinx_version_parts[1] < 6:
+    html_use_smartypants = False
+else:
+    smartquotes = False
 
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}


doc/sphinx/conf.py
```diff
@@ -140,7 +140,10 @@ def setup(app):
 
 # If true, SmartyPants will be used to convert quotes and dashes to
 # typographically correct entities.
-#html_use_smartypants = True
+# Better to disable this to prevent alteration of command-line arguments
+# Sphinx <1.6 use `html_use_smartypants` and >=1.6 use `smartquotes`
+html_use_smartypants = False
```
volans-: In another project to be compatible with both older and newer versions of Sphinx I've used:

```suggestion
from sphinx import __version__ as sphinx_version

sphinx_version_parts = [int(i) for i in sphinx_version.split('.')]
if sphinx_version_parts[0] == 1 and sphinx_version_parts[1] < 6:
    html_use_smartypants = False
```

I didn't had to set `smartquotes` to `False`, but I use sphinx-argparse for documenting the CLI arguments.

My 2 cents

xdelaruelle: Thanks for the suggestion, I have rewritten the commit to make use of the test code you propose.

thiell:
@volans- do you want to give your approval before we merge this? thx

doc/sphinx/conf.py
```diff
@@ -140,7 +140,14 @@ def setup(app):
...
 # typographically correct entities.
-#html_use_smartypants = True
+# Better to disable this to prevent alteration of command-line arguments
+# Sphinx <1.6 use `html_use_smartypants` and >=1.6 use `smartquotes`
+from sphinx import __version__ as sphinx_version
+sphinx_version_parts = [int(i) for i in sphinx_version.split('.')]
+if sphinx_version_parts[0] <= 1 and sphinx_version_parts[1] < 6:
+    html_use_smartypants = False
+else:
+    smartquotes = False
```
volans-: It would be nice if CI was generating the documentation too so that we could see the result before merging. But also easier to adapt in case of issues.

doc/sphinx/conf.py
```diff
@@ -140,7 +140,14 @@ def setup(app):
 
 # If true, SmartyPants will be used to convert quotes and dashes to
 # typographically correct entities.
-#html_use_smartypants = True
+# Better to disable this to prevent alteration of command-line arguments
+# Sphinx <1.6 use `html_use_smartypants` and >=1.6 use `smartquotes`
+from sphinx import __version__ as sphinx_version
```
volans-: [nitpick] I would have put the import at the top of the file together with the others, but YMMV

doc/sphinx/conf.py
thiell: True that!! I'll see what I can do.

PR #405, opened 2019-05-21
Slurm bindings

kcgthb:
Fix issue with the `slurmuser` binding, that was using job ids instead of user names.


degremont:
Thanks for the patches. For review readability, could you rebase your branch on top of master (and not a merge) and force-push your PR to update it? Thanks

PR #409, opened 2019-07-18
doc: fix typo, %r is not for rank

fihuer:
Hi,
Here is a small PR for a doc fix.

'man clush' says that '%r'  is correct for the exec worker. It should be '%n'.

Cheers,

Romain Fihue


degremont:
Thanks a lot for this patch.

Unfortunately, the doc is not automatically built. So, could you fix this also in man page in `doc/man/man1/clush.1` in the same patch?

fihuer:
> Thanks a lot for this patch.
> 
> Unfortunately, the doc is not automatically built. So, could you fix this also in man page in `doc/man/man1/clush.1` in the same patch?

Done

PR #422, opened 2019-11-17
Clarify NodeSet docs on functions that do not return a NodeSet

mattaezell:
The `*_update` functions operate on a NodeSet in-place and return `None`.  Make this clearer so someone doesn't attempt to do `s = s.intersection_update(t)`


thiell:
Thanks @mattaezell!

Issue #428, opened 2020-01-30
CLI_COLOR support


kcgthb:
Hi!

To build on #334 and #390, would it be possible to add support for `NO_COLOR` (https://no-color.org/) in ClusterShell?

degremont:
I think CLI_COLOR would have more value there (https://bixense.com/clicolors/)

kcgthb:
What about supporting both?

dupgit:
Hi,
I propose PR #432 to solve this Issue. It implements both NO_COLOR and CLI_COLOR that has different purposes.
Regards,
Olivier.

PR #432, opened 2020-03-17
Add support for NO_COLOR and CLICOLOR (#428)

dupgit:
Takes care of NO_COLOR, CLICOLOR and CLICOLOR_FORCE environment variables. NO_COLOR takes precedence over CLICOLOR_FORCE itself taking precedence over CLICOLOR itself taking precedence over --color command line option.

When none of NO_COLOR, CLICOLOR and CLICOLOR_FORCE environment variables
are defined the old behavior still works and command line options are taken into account.


lib/ClusterShell/CLI/Display.py
```diff
@@ -137,6 +151,42 @@ def __init__(self, options, config=None, color=None):
...
+            return False
+        else:
+            # CLICOLOR is not set
+            return None
+
+    def _has_no_color(self):
+        """Tests NO_COLOR environment variable to determine if color
+        must be muted on output."""
+
+        if os.getenv('NO_COLOR') is not None:
```
degremont: prefer

    return 'NO_COLOR' in os.environ

And this could be directly use above maybe, no need for a function?

lib/ClusterShell/CLI/Display.py
```diff
@@ -137,6 +151,42 @@ def __init__(self, options, config=None, color=None):
...
+        use color or not on output.
+        """
+        # When CLICOLOR_FORCE is set to something else than 0
+        # colors must be used.
+        if os.getenv("CLICOLOR_FORCE", "0") != "0":
+            return True
+
+        cli_color = os.getenv("CLICOLOR")
+
+        if cli_color is None:
```
degremont: Looks like:

    if cli_color is None:
        return None
    elif cli_color != "0":
        return sys.stdout.isatty()
    else:
        return False

lib/ClusterShell/CLI/Display.py
```diff
@@ -164,28 +162,12 @@ def _has_cli_color(self):
 
         if cli_color is None:
             return None
-        elif cli_color != "0" and sys.stdout.isatty():
-            # CLICOLOR are ment to be used
-            return True
         elif cli_color != "0":
-            # CLICOLOR should be used but stdout is not a tty.
-            return False
+            # CLICOLOR are ment to be used if stdout is a tty
```
degremont: not sure what 'ment' means?

lib/ClusterShell/CLI/Display.py
```diff
@@ -164,28 +162,12 @@ def _has_cli_color(self):
...
+            return sys.stdout.isatty()
         elif cli_color == "0":
```
degremont: prefer a simple `else` here

degremont:
Could be good to change the commit message.

- Squash all your commits in one
- Add a small subject line as first commit line, starting with `Defaults: `
- Add `(#428)` at the end
- Add `Closes #428` in the commit message
- Add a Signed-Off line

See how other commit messages look like.

degremont:
Looks good to me. I was wondering if the doc text should not be written using a dedicated section for Environment variables, or something similar.

I will let @thiell validate this part.

thiell:
Hi @dupgit,
Thanks much for this PR. We think the command line option `--color` should always have precedence over these environment variables. It makes more sense especially if a user wants to be sure to disable any colors with `--color=never`. But actually, in all cases when `--color` is specified by the user, we think that environment variables should just be ignored (even for `--color=auto`).
Could you please make this last change?
Again, thanks much!

doc/txt/clubak.txt
```diff
@@ -57,7 +57,7 @@ OPTIONS
...
 -F, --fast     faster but memory hungry mode (preload all messages per node)
 -T, --tree     message tree trace mode; switch to enable ``ClusterShell.MsgTree`` trace mode, all keys/nodes being kept for each message element of the tree, thus allowing special output gathering
---color=WHENCOLOR   whether to use ANSI colors to surround node or nodeset prefix/header with escape sequences to display them in color on the terminal. *WHENCOLOR* is ``never``, ``always`` or ``auto`` (which use color if standard output refers to a terminal). Color is set to [34m (blue foreground text) and cannot be modified.
+--color=WHENCOLOR   ``clush`` is compliant with NO_COLOR, CLICOLOR and CLICOLOR_FORCE environment variables. NO_COLOR takes precedence over CLICOLOR_FORCE which takes precedence over CLICOLOR. If these environment variables are not used then ``--color`` tells whether to use ANSI colors to surround node or nodeset prefix/header with escape sequences to display them in color on the terminal. *WHENCOLOR* is ``never``, ``always`` or ``auto`` (which use color if standard output refers to a terminal). Color is set to [34m (blue foreground text) and cannot be modified.
```
thiell: --color should have precedence over environment variables

thiell:
Hi @dupgit,

Thanks for the change and sorry for the delay. While testing your patch, I found an issue that annoys me a bit. If `color` is defined in `clush.conf`, the color environment variables don't seem to be taken into account anymore, even if `color: auto` in clush.conf.

We could fix the patch so that if `color: auto` is defined, environment variables are still considered. They probably should not if  `color: always` or `color: never` is set.

Or perhaps even better, I'm wondering if we should offer more choices in `clush.conf`: something like `env` (use env variables, or otherwise auto), `auto`, `always`, `never` and set `env` as the new default. Let me know what you think.

dupgit:
> Hi @dupgit,

Hi @thiell,

> Thanks for the change and sorry for the delay. While testing your patch, I found an issue that annoys me a bit. If `color` is defined in `clush.conf`, the color environment variables don't seem to be taken into account anymore, even if `color: auto` in clush.conf.

Thanks for the rewiew. I thought that options that are set in either command line or configuration file should take precedence over color environment variables.

> We could fix the patch so that if `color: auto` is defined, environment variables are still considered. They probably should not if `color: always` or `color: never` is set.

I will fix the proposed patch in that way: environment variable are still taken into account when color:auto or no color option is configured in `clush.conf` file. Nothing changes for command line: if any parameter is set in the command line environment variables are ignored.

> Or perhaps even better, I'm wondering if we should offer more choices in `clush.conf`: something like `env` (use env variables, or otherwise auto), `auto`, `always`, `never` and set `env` as the new default. Let me know what you think.

I think that this will be more confusing to the user and will make the patch more complex. Getting `auto` taking environment variable automatically into account is simpler to me: if `env` is defined and no environment variable is defined at all then the expected behavior would be like `auto`. So lets keep it and add the fact that it can now take environment variable into account.

dupgit:
@thiell @degremont is there anything left here to help merging this patch ?

thiell (approved):
Thank you, @dupgit, for your contribution!

PR #434, opened 2020-03-25
CLI/Clush: Use the program name in OutputHandlers and RunTimer

### LINKED ISSUE #433: Use clush OutputHandlers in external tools
`clush` has a rich set of output handlers defined in `Clush.py`.  Instead of reimplement them for an external tool, it would be nice to be able to use them directly.  Unfortunately, the error messages currently hard-code `clush:` as a prefix.

I would propose making the prefix configurable, and potentially moving the `OutputHandler`s into either `Display.py` or their own new file.

I'm happy to implement and open a PR, but I wanted to get some consensus before starting work.  Thoughts?

mattaezell:
Also allow manually-instantiated OutputHandlers to specify an alternate program
name.

Closes #433

Signed-off-by: Matt Ezell <ezellma@ornl.gov>

### DIFF
--- a/lib/ClusterShell/CLI/Clush.py
+++ b/lib/ClusterShell/CLI/Clush.py
@@ -82,15 +82,16 @@ class StdInputHandler(EventHandler):
         self.master_worker.write(msg)
 
 class OutputHandler(EventHandler):
-    """Base class for clush output handlers."""
+    """Base class for generic output handlers."""
 
-    def __init__(self):
+    def __init__(self, prog=None):
         EventHandler.__init__(self)
         self._runtimer = None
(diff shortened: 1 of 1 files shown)


mattaezell:
I debated whether or not to manually specify `prog='clush'` in the `clush` invocation of the output handlers or let it use `sys.argv[0]` by default.  People using the command as `clush3` will now see `clush3` in their error messages (the unit tests don't change because they specify argv manually).

I also didn't add a specific test for this change, as the tests really don't belong in `CLIClush.py`.  But happy to look into adding one if you have any guidance.

mattaezell:
@thiell mind taking a look at this one? Thanks.

thiell:
Looks good, thanks!

PR #448, opened 2020-09-16
Return maxrc properly for Rsh Worker

hawartens:
The current version of the Rsh Worker does not properly return
the max error code with the -S option. This commit adds in a little
"magic" to the end of the rsh command to pass back the return code
when requested. Basically works exactly as pdsh does to extract the
return code.

### DIFF
--- a/lib/ClusterShell/Worker/Rsh.py
+++ b/lib/ClusterShell/Worker/Rsh.py
@@ -26,6 +26,7 @@ This is also the base class for rsh evolutions, like Ssh worker.
 
 import os
 import shlex
+import re
 
 from ClusterShell.Worker.Exec import ExecClient, CopyClient, ExecWorker
 
@@ -59,7 +60,24 @@ class RshClient(ExecClient):
         cmd_l.append("%s" % self.key)  # key is the node
         cmd_l.append("%s" % self.command)
 
+        # rsh does not properly return exit status
+        # force the exit status to be printed out
+        cmd_l.append("; echo XXRETCODE: $?")
+
         return (cmd_l, None)
+
+    def _on_nodeset_msgline(self, nodes, msg, sname):
+        """local wrapper over _on_node_msgline that can also handle nodeset"""
+        match = re.search("^XXRETCODE: (\d+)$", msg.decode("utf-8"))
+        if match:
+            self.rsh_rc = int(match.group(1))
+        self.worker._on_node_msgline(nodes, msg, sname)
+
+    def _on_nodeset_close(self, nodes, rc):
+        """local wrapper over _on_node_rc that can also handle nodeset"""
+        if (rc == 0 or rc == 1) and self.rsh_rc is not None:
+            rc = self.rsh_rc
+        self.worker._on_node_close(nodes, rc)
 
 
 class RcpClient(CopyClient):
@@ -122,7 +140,7 @@ class WorkerRsh(ExecWorker):
     Remote Copy (rcp) usage example:
        >>> worker = WorkerRsh(nodeset, handler=MyEventHandler(),
        ...                     source="/etc/my.conf",
-       ...                     dest="/etc/my.conf")
+       ...                     dest="/etc/my.conf.bak")
        >>> task.schedule(worker)      # schedule worker for execution
        >>> task.resume()              # run
 
--- a/lib/ClusterShell/Worker/Rsh.py
+++ b/lib/ClusterShell/Worker/Rsh.py
@@ -74,7 +74,7 @@ class RshClient(ExecClient):
 
     def _on_nodeset_msgline(self, nodes, msg, sname):
         """Override _on_nodeset_msgline to parse magic return code"""
-        match = re.search("^XXRETCODE: (\d+)$", msg.decode("utf-8"))
+        match = re.search("^XXRETCODE: (\d+)$", msg.decode())
         if match:
             self.rsh_rc = int(match.group(1))
         else:
--- a/lib/ClusterShell/Worker/Rsh.py
+++ b/lib/ClusterShell/Worker/Rsh.py
@@ -26,7 +26,6 @@ This is also the base class for rsh evolutions, like Ssh worker.
 
 import os
 import shlex
-import re
 
 from ClusterShell.Worker.Exec import ExecClient, CopyClient, ExecWorker
 
@@ -46,7 +45,6 @@ class RshClient(ExecClient):
         path = task.info("rsh_path") or "rsh"
         user = task.info("rsh_user")
         options = task.info("rsh_options")
-        maxrc = task.info("maxrc", False)
 
         cmd_l = [os.path.expanduser(pathc) for pathc in shlex.split(path)]
 
@@ -61,30 +59,7 @@ class RshClient(ExecClient):
         cmd_l.append("%s" % self.key)  # key is the node
         cmd_l.append("%s" % self.command)
 
-        # Save the return code from the command if required
-        if maxrc:
-            cmd_l.append("; echo XXRETCODE: $?")
-
         return (cmd_l, None)
-
-    def _parse_line(self, line, sname):
-
-        # Read the return code from the command if required
-        task = self.worker.task
-        maxrc = task.info("maxrc", False)
-        if maxrc:
-            match = re.search("^XXRETCODE: (\d+)$", line.decode("utf-8"))
-            if match:
-                self.worker.current_rc = int(match.group(1))
-            else:
-                self.worker._on_node_msgline(self.key, line, sname)
-        else:
-            self.worker._on_node_msgline(self.key, line, sname)
-
-    def _handle_read(self, sname):
-        """Engine is telling us a read is available."""
-        for msg in self._readlines(sname):
-            self._parse_line(msg, sname)
 
 
 class RcpClient(CopyClient):
@@ -147,17 +122,12 @@ class WorkerRsh(ExecWorker):
     Remote Copy (rcp) usage example:
        >>> worker = WorkerRsh(nodeset, handler=MyEventHandler(),
        ...                     source="/etc/my.conf",
-       ...                     dest="/etc/my.conf.bak")
+       ...                     dest="/etc/my.conf")
        >>> task.schedule(worker)      # schedule worker for execution
        >>> task.resume()              # run
 
     connect_timeout option is ignored by this worker.
     """
-
-    def _on_node_close(self, node, rc):
-        if rc is None or self.current_rc > rc:
-            rc = self.current_rc
-        ExecWorker._on_close(self, node, rc)
 
     SHELL_CLASS = RshClient
     COPY_CLASS = RcpClient
--- a/lib/ClusterShell/Worker/Rsh.py
+++ b/lib/ClusterShell/Worker/Rsh.py
@@ -26,6 +26,7 @@ This is also the base class for rsh evolutions, like Ssh worker.
 
 import os
 import shlex
+import re
 
 from ClusterShell.Worker.Exec import ExecClient, CopyClient, ExecWorker
 
@@ -45,6 +46,7 @@ class RshClient(ExecClient):
         path = task.info("rsh_path") or "rsh"
         user = task.info("rsh_user")
         options = task.info("rsh_options")
+        maxrc = task.info("maxrc", False)
 
         cmd_l = [os.path.expanduser(pathc) for pathc in shlex.split(path)]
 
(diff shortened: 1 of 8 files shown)


degremont:
Thanks @hawartens for your PR.

I'm seeing that your code change is actually adding this little "magic" and also adding `maxrc` as new `clush.conf` option. This is probably worthing adding in the commit or make 2 commits of that.

I'm not against adding this option in config file is you think this is useful, but it seems you did this to propagate the `maxrc` flag value down to the Worker code. If this is just for this, you can simply add your `echo` command all the time whatever the `maxrc` value is.
The problem is I don't understand why maxrc is not working for Rsh. It looks to me `rsh` is a simple command working like any command runnable with an ExecWorker or like a SshWorker.
Could you explain to me (and maybe in the commit message) what was the problem you are solving with this code?

Checking the code, that would means `ExecClient._close()` is not working as expected for `rsh`. Could you clarify what you are seeing?

hawartens:
Thanks @degremont. Correct I added the maxrc option so that I could propagate it nicely down to the worker code a little more cleanly. I wanted the code to behave more like pdsh in this respect and not add the magic unless it was necessary. It is possible to grab maxrc in the Worker by looking at self.worker.eh._display.maxrc but it felt that would be violating your intent. If you prefer, I can package this as two commits (I did it together for the exact reason you surmised).

degremont:
Understood for the `maxrc` config change. But what about the original problem?
Could you explain in details what is not working with `RshWorker` which is working with `ssh` or `exec` ?

hawartens:
> Understood for the `maxrc` config change. But what about the original problem?
> Could you explain in details what is not working with `RshWorker` which is working with `ssh` or `exec` ?

Sure.  `rsh` does not correctly set return codes. As far as I know, it never has. See example below:

```
> cat /tmp/foo.sh
#!/bin/bash
exit 1

> ssh localhost /tmp/foo.sh
> echo $?
1

> rsh localhost /tmp/foo.sh
> echo $?
0
```

This manage shows the reason:
https://docs.oracle.com/cd/E36784_01/html/E36870/rsh-1.html
```
rsh does not return the exit status code of command.
```

degremont:
Oh! I understood now how this trick makes sense. 
But, in this case, why not simply apply this trick all the time and just update the local rc for each rsh command, as this is done for ssh or exec?
that way, the -S will be automatically handled.

hawartens:
> Oh! I understood now how this trick makes sense.
> But, in this case, why not simply apply this trick all the time and just update the local rc for each rsh command, as this is done for ssh or exec?
> that way, the -S will be automatically handled.


I patterned it after the way it works in Pdsh.  We only do the "magic" if the user requests it.  This way we do not do any extra work if it is not necessary.  If you would prefer to have it done all the time, that is okay with me, was just trying to avoid doing it unless it was specifically asked for.

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -59,8 +61,31 @@ def _build_cmd(self):
         cmd_l.append("%s" % self.key)  # key is the node
         cmd_l.append("%s" % self.command)
 
+        # Save the return code from the command if required
+        if maxrc:
+            cmd_l.append("; echo XXRETCODE: $?")
```
thiell, line 66: If I understand correctly, but let me know if I missed something, what you rather want is to add `"; echo XXRETCODE: $?"` to `self.command` itself. Otherwise, you will get the return code of rsh.

and this should be part of the command string, likely by doing something like the following, before appending self.command to cmd_l:

```
self.command += "; echo XXRETCODE: $?"
```

thiell (changes requested):
Thanks for your PR. I would prefer to avoid adding a new clush option maxrc which would be confusing, and rather fix the max return code with the Rsh worker. Users of library could also benefit from this when the Rsh worker is used, and thus get proper return codes in that case. I think the overhead is minimal. Let us know if you can update your PR in that sense.

lib/ClusterShell/Worker/Rsh.py
hawartens: Hello @thiell,
Perhaps I am confused.  I do not believe I really added a new Clush option.  The -S option already exists in Clush, but it did not work correctly for rsh.  All I did really was to pass that down to the Rsh Worker so that it would know when to do the "magic" since I was trying to emulate Pdsh and not do the "echo XXRETCODE: $?" if it was not required.
I did externalize the -S option as "maxrc" in the configuration file so that it would be possible to avoid passing -S on the command line on every invocation if you want that to be the default.
I am happy to submit that in multiple pull requests if you prefer. So I have a couple questions:

1) Would you prefer appending "; echo XXRETCODE: $?" for every command to the rsh worker (whether -S is passed in or not)? Happy to do it that way if you want, was just trying to avoid running the extra echo command if not requested.

2) Do you dislike having the ability to set the -S in the configuration file (using "maxrc")?  If you don't want that ability at all I am happy to remove it, leave it here, or commit it as a separate pull request.

Thanks!

thiell: Sorry I meant a new clush option (maxrc) in the configuration file clush.conf. I understand your logic and we could indeed add `maxrc` to `clush.conf` to enable `clush -S` by default. However, I don't like the use of `task.set_info()` to pass this information to the worker. At the end, I think maxrc should stay a clush-only thing and should not impact the library/workers.

So I guess:

1. yes, definitely I would prefer that we end up with a proper rsh worker that always processes the return code. It's the same with the pdsh worker, we always gather command return codes, even if clush -S is not used. Clush then can decide if it wants to use maxrc or not, but at least we would have fixed the worker's behavior. I really thing the overhead is minimal as `echo $?` is usually a shell built-in command.

2. thanks for your flexibility! I think it's best to put the maxrc change into a different PR, so we can discuss about it separately.

If you're ok with the above, could you please update this PR so that it will focus on fixing the Rsh worker to handle command return codes, and open a separate PR to add maxrc to clush.conf that will enable `-S` by default with clush. Thanks again!

hawartens: Sure!  No problem!

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -59,8 +61,31 @@ def _build_cmd(self):
...
+        if maxrc:
+            match = re.search("^XXRETCODE: (\d+)$", line.decode("utf-8"))
+            if match:
+                self.worker.current_rc = int(match.group(1))
+            else:
+                self.worker._on_node_msgline(self.key, line, sname)
+        else:
+            self.worker._on_node_msgline(self.key, line, sname)
+
+    def _handle_read(self, sname):
```
degremont, line 84: I think you should override `_on_nodeset_msgline()` instead, because this is called from `_handle_read()` and `_flush_read()` and you are missing `_flush_read()` right now.

You just need to search for the pattern, then, call `_on_nodeset_msgline()`. Store the rc in a client local variable. Then override `_on_nodeset_close()` and use that value.
This function could be called with either the actual rc from rsh (0 or 1) or 128 - SIGNUM is the rsh process was killed by a signal. You should probably check for 0 or 1 and use the saved rc only in this case. Mmh... actually, the local rc would be None in this case, so that should not be difficult anyway...

degremont:
I add some coding guidelines for this feature. I'm inline with @thiell and limit this PR to the rc feature. Don't include maxrc change here.

lib/ClusterShell/Worker/Rsh.py
hawartens: Thanks @degremont.  I believe I have made your requested changes, though I was a little confused when you said I should override `_on_nodeset_msgline()` and pithing there I should call `_on_nodeset_msgline ()`.  I am not 100% sure, but I believe you meant that I should call `_on_node_msgline()` from within the overridden function.  Please let me know if I misunderstood.  Thanks!

lib/ClusterShell/Worker/Rsh.py
hawartens: @thiell.  Thanks.  I believe that I have implemented 1.  I believe that I did what @degremont requested as well.

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -59,8 +60,25 @@ def _build_cmd(self):
...
+        cmd_l.append("; echo XXRETCODE: $?")
+
         return (cmd_l, None)
 
+    def _on_nodeset_msgline(self, nodes, msg, sname):
+        """local wrapper over _on_node_msgline that can also handle nodeset"""
+        match = re.search("^XXRETCODE: (\d+)$", msg.decode("utf-8"))
+        if match:
+            self.rsh_rc = int(match.group(1))
+        self.worker._on_node_msgline(nodes, msg, sname)
```
degremont, line 74: add a `else` above, unless you will call _on_node_msgline for output like` XXRETCODE:` and this will be visible by users.

Also, call the _on_node_msgline() from the parent class here:

    ExecClient._on_nodeset_msgline(self, nodes, msg, sname)

That's important because, as you can see in this function, there is a special logic there.

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -59,8 +60,25 @@ def _build_cmd(self):
...
+
+    def _on_nodeset_close(self, nodes, rc):
+        """local wrapper over _on_node_rc that can also handle nodeset"""
+        if (rc == 0 or rc == 1) and self.rsh_rc is not None:
+            rc = self.rsh_rc
+        self.worker._on_node_close(nodes, rc)
```
degremont, line 80: same here, use parent's method

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -122,7 +140,7 @@ class WorkerRsh(ExecWorker):
     Remote Copy (rcp) usage example:
        >>> worker = WorkerRsh(nodeset, handler=MyEventHandler(),
        ...                     source="/etc/my.conf",
-       ...                     dest="/etc/my.conf")
+       ...                     dest="/etc/my.conf.bak")
```
degremont, line 143: why this change?

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -59,8 +60,25 @@ def _build_cmd(self):
...
 
+        # rsh does not properly return exit status
+        # force the exit status to be printed out
+        cmd_l.append("; echo XXRETCODE: $?")
+
         return (cmd_l, None)
 
+    def _on_nodeset_msgline(self, nodes, msg, sname):
+        """local wrapper over _on_node_msgline that can also handle nodeset"""
+        match = re.search("^XXRETCODE: (\d+)$", msg.decode("utf-8"))
```
degremont, line 71: utf8 handling is a pain :) @thiell thoughts on that part?

degremont (changes requested):
Thanks for the update! The principles are what I was looking for. I requested small changes.

lib/ClusterShell/Worker/Rsh.py
hawartens: Just noticed an error in the comment I believe. Did not think the source and destination should be the same. Had that change in the previous pull request.

degremont: I don't think there is an error error. Rcp is for copying files from 1 host to another one. Let's imagine you want to copy `/etc/passwd` from `host1` to `host[2-10]` to have the same file everywhere? The source and destination will be the same file name (but not the same file).

hawartens: Fair point.

lib/ClusterShell/Worker/Rsh.py
hawartens: Done!

lib/ClusterShell/Worker/Rsh.py
hawartens: Removed!

lib/ClusterShell/Worker/Rsh.py
hawartens: Done!

lib/ClusterShell/Worker/Rsh.py
```diff
@@ -59,8 +60,31 @@ def _build_cmd(self):
...
+        # rsh does not properly return exit status
+        # force the exit status to be printed out
+        cmd_l.append("; echo XXRETCODE: $?")
+
         return (cmd_l, None)
 
+    def _on_nodeset_start(self, nodes):
+        """Override _on_nodeset_start to guarantee that rsh_rc is initialized """
+        self.rsh_rc = None
+        ExecClient._on_nodeset_start(self, nodes)
```
degremont, line 72: Usually attribute declaration should be done in `__init__()`. That's ok like that, but, if you happen to refresh this code, please move that to __init__

degremont (approved):
Overall I'm ok. Just a minor comment. I will let @thiell double check, especially the utf thing.

lib/ClusterShell/Worker/Rsh.py
hawartens: Agree.  Just seemed like overkill to override init().  In this case it seemed cleaner to do this, but I will move it to __init__(), since it really is the right thing to do!

hawartens: Done!

lib/ClusterShell/Worker/Rsh.py
thiell: I would rather use `decode()` without argument, like in the [Pdsh worker](https://github.com/cea-hpc/clustershell/blob/master/lib/ClusterShell/Worker/Pdsh.py#L139). Default encoding is `'utf-8'` anyway. Note that by default the `errors` argument of `decode()` is set to `'strict'` which may raise a UnicodeError exception, but that is fine with me here as the output text encoding after XXRETCODE is expected.

thiell (changes requested):
Looks very good overall. I would just fix the decode() method to avoid the explicit utf-8 encoding. Thanks so much.

lib/ClusterShell/Worker/Rsh.py
hawartens: Done.  Thanks!

thiell (approved):

PR #451, opened 2020-09-22
Add maxrc Option to clush.conf

hawartens:
Externalize the -S option as "maxrc" in the configuration file to
avoid passing it on the command line on every invocation.


degremont (approved):
Looks good. Just wondering if `--maxrc` is needed as we cannot use a true/false flag.

PR #460, opened 2021-03-05
Avoid python3 error with no stdin

mattaezell:
Under python3, `sys.stdin` is `None` if there is no stdin attached to the process:

```
$ python3 -c 'import sys; assert not sys.stdin.isatty()' 0<&-
  File "<string>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'isatty'
```

Originally reported at https://bugs.schedmd.com/show_bug.cgi?id=11021


thiell:
Great finding @mattaezell! I will review your patch ASAP. We should include the fix in 1.8.4.

thiell:
It's interesting. I can't find any documentation to explain when `sys.stdin` can be None in the Python documentation. Looks like other developers are facing the same problem with Python 3: https://stackoverflow.com/questions/17458728/when-is-sys-stdin-none-in-python/17459258
I guess we have no choice but checking whether it is defined, like you're doing in your patch indeed. We'll need to find a way to make a test too.

thiell:
It's set to None here:
https://hg.python.org/cpython/file/dbceba88b96e/Python/pythonrun.c#l1201
```
    /* Under some conditions stdin, stdout and stderr may not be connected
     * and fileno() may point to an invalid file descriptor. For example
     * GUI apps don't have valid standard streams by default.
     */
```

thiell:
Ah! It is actually documented under `sys.__stdin__` in the [Python 3 documentation](https://docs.python.org/3/library/sys.html#sys.__stdin__). It was not documented in [Python 2](https://docs.python.org/2/library/sys.html#sys.stdin).

_Note: Under some conditions stdin, stdout and stderr as well as the original values __stdin__, __stdout__ and __stderr__ can be None. It is usually the case for Windows GUI apps that aren’t connected to a console and Python apps started with pythonw._

lib/ClusterShell/CLI/Clush.py
```diff
@@ -966,7 +966,8 @@ def main():
     task.set_default("USER_handle_SIGUSR1", user_interaction)
 
     task.excepthook = sys.excepthook
-    task.set_default("USER_stdin_worker", not (sys.stdin.isatty() or \
+    task.set_default("USER_stdin_worker", not ((sys.stdin is not None and \
```
mattaezell, line 969: The original reporter reported that this patch does not fix the issue. There's another access of sys.stdin in an area of the code that I thought shouldn't be reached. Looking at this code, I've either confused myself, or the original code was incorrect.  If stdin is a tty, it was setting this to false - is that the correct behavior?

thiell: There are many cases here that is a bit confusing. Let me explain some of the possible conditions:

TTY stdin: If stdin is connected to a tty, `USER_stdin_worker` is set to False, and `bind_stdin()` is not called, but user_interaction might be set to True (we also check for stdou) to enable terminal interaction (handled by `ttyloop()`).


non-TTY stdin: `bind_stdin()`  is called when a (valid) non-tty stdin is used and also when `--nostdin` is not specified. It creates a thread that is dedicated to read stdin and push the data to a "task port" using a handler `StdInputHandler()`.

the default "USER_stdin_worker" controls whether bind_stdin() is called or not.

This is used for example when we do `echo "foo" | clush -w ... cmd`

If `sys.stdin` is None, `USER_stdin_worker` should be set to False so that bind_stdin is not called. Your patch set `USER_stdin_worker` to True if sys.stdin is None. We want the opposite.

lib/ClusterShell/Gateway.py
```diff
@@ -343,7 +343,7 @@ def gateway_main():
     task.set_default("stdout_msgtree", False)
     task.set_default("stderr_msgtree", False)
 
-    if sys.stdin.isatty():
+    if sys.stdin is not None and sys.stdin.isatty():
```
thiell, line 346: I would prefer to split this into two separate conditions, and update the error message accordingly in case sys.stdin is None.

lib/ClusterShell/CLI/Clush.py
```diff
@@ -635,7 +635,7 @@ def bind_stdin(worker, display):
     """Create a stdin->port->worker binding: connect specified worker
     to stdin with the help of a reader thread and a ClusterShell Port
     object."""
-    assert not sys.stdin.isatty()
+    assert sys.stdin is None or not sys.stdin.isatty()
```
thiell, line 638: (defect) we want to make sure sys.stdin is not None in `bind_stdin()`

thiell (changes requested):

lib/ClusterShell/Gateway.py
mattaezell: Is the gateway code guaranteed to always have stdin?

mattaezell: I see it uses it later, so seems like a fair assumption.

lib/ClusterShell/CLI/Clush.py
mattaezell: It makes sense to me now. Thanks for the explanation.

lib/ClusterShell/Gateway.py
thiell: Good point. The gateway module is launched by clustershell itself and stdin is always connected, so it's very unlikely that sys.stdin would be None. So maybe there is nothing to change here after all. if sys.stdin is None here, it will fail anyway and we can't do anything about it. So maybe do nothing?

lib/ClusterShell/Gateway.py
```diff
@@ -343,6 +343,10 @@ def gateway_main():
     task.set_default("stdout_msgtree", False)
     task.set_default("stderr_msgtree", False)
 
+    if sys.stdin is None:
```
thiell, line 346: `sys.stdin` is used earlier above, so that actually won't work. Either put the check above, or leave the Gateway as is for now (I'm fine either way).

thiell (changes requested):
looks better :)

PR #462, opened 2021-08-18
Worker/Ssh: support %h %n variables

valuefish:


thiell:
Thanks for your pull request. For backward compatibility, if added to a minor release, it would be better if this feature is disabled by default, so we don't change existing behavior with commands using %n for example and not expecting to be resolved. We need a switch to enable/disable it, I think. This could perhaps be added to defaults.conf.

PR #468, opened 2022-02-08
NodeSet: add special notation @@source to expand group names

thiell:
This change allows access to all group names of a specific source in
node set expressions. The special operator `@@` followed by the source
expands group names that can then be manipulated as a set.

Note: In that case, group names are not prefixed with @ and as such, not
further expanded into nodes by NodeSet.

### DIFF
--- a/lib/ClusterShell/NodeSet.py
+++ b/lib/ClusterShell/NodeSet.py
@@ -916,6 +916,8 @@ class ParsingEngine(object):
             namespace, group = grpstr.split(':', 1)
         if group == '*': # @* or @source:* magic
             reslist = self.all_nodes(namespace)
+        elif group.startswith('@'): # @@source group name list
+            reslist = self.grouplist(grpstr[1:])
         else:
             reslist = self.group_resolver.group_nodes(group, namespace)
         return ','.join(reslist), namespace
--- a/doc/man/man1/cluset.1
+++ b/doc/man/man1/cluset.1
@@ -1,6 +1,6 @@
 .\" Man page generated from reStructuredText.
 .
-.TH CLUSET 1 "2019-12-01" "1.8.3" "ClusterShell User Manual"
+.TH CLUSET 1 "2022-03-19" "1.8.4" "ClusterShell User Manual"
 .SH NAME
 cluset \- compute advanced cluster node set operations
 .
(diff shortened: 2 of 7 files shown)


degremont:
Interesting feature! I'm trying to have this match into the other feature we have for NodeSet syntax/arithmetic.

So far, we supports various syntaxes to identify node name, or group names that will be mapped into node names.
Group name related syntax refers to the current default source by default. An additional prefix could be added to point to a different source.

I like the idea of fetching the group list with a nodeset syntax, but I would like it to have a syntax that could be pointing to the current source without knowing its name. 

* Syntax <xxxx> to get the group name of current source
* Syntax <yyyy> to get the group name for source B. Ideally this syntax is the same than the other, where we just add the source name prefix `source:`.

thiell:
Thanks! This patch implements your second point. Note that when listing group names from a source, only the source name is needed. For example, with this patch, `@@B` will get you the group names for source B.

With the `@source:group` syntax, the input parameter is mainly the group name.  With `@@` however, it's the source itself. I can propose to see if we can implement`@@` (not followed by an explicit source name) what would resolve group names of the default/current source.

Lastly, but I doubt it will be very useful in practice, we could implement `@@source:<group_pattern>`, to filter the group names based on the wildcard pattern group_pattern, like `@@B:pdu*`.

Both last ideas are compatible with this patch. Let me know what you think.

kcgthb:
I like it the way it is right now, not only because the patch is just so tiny and elegant (which talks to the quality of the existing code), but also because I feel it's more intuitive for users.

Listing the group names in group source `foo` with `cluset -e @@foo` seems pretty easy to grasp.

I'm not sure implementing a `@@source:<group_pattern>` filter is worth the trouble, it's a job better left of for `grep` and friends, IMHO.

And as for listing the groups in the current source without knowing its name, it works out the box already with:
```
cluset -e @@
```
Did I mention that patch was awesome? ;)

lib/ClusterShell/NodeSet.py
```diff
@@ -916,6 +916,8 @@ def parse_group_string(self, nodegroup, namespace=None):
             namespace, group = grpstr.split(':', 1)
         if group == '*': # @* or @source:* magic
             reslist = self.all_nodes(namespace)
+        elif group.startswith('@'): # @@source group name list
+            reslist = self.grouplist(grpstr[1:])
```
degremont, line 920: ok, no source name is supported because `grpstr[1:]` will be expanded to `''` and we are testing `namespace` as a boolean, so both `None` (official value)  and `''` will evaluate to `False`.

degremont (changes requested):
Ok, if you both think `@@` is a clear enough syntax, i'm not against it. Just make that clear in the doc and the tests that this is an officially supported syntax.

Agreed that name filtering could be added later, if people asks for it.

thiell:
> Just make that clear in the doc and the tests that this is an officially supported syntax.

done!

PR #469, opened 2022-02-09
slurm.conf.example: Filter out more Slurm node state flags

kcgthb:
Slurm 21.08 introduced a new node state flag (^), and we missed a few before.
See https://slurm.schedmd.com/sinfo.html#SECTION_NODE-STATE-CODES for a complete list

### DIFF
--- a/conf/groups.conf.d/slurm.conf.example
+++ b/conf/groups.conf.d/slurm.conf.example
@@ -18,8 +18,8 @@ reverse: sinfo -h -N -o "%R" -n $NODE
 [slurmstate,st]
 map: sinfo -h -o "%N" -t $GROUP
 all: sinfo -h -o "%N"
-list: sinfo -h -o "%T" | tr -d '*~#$@+'
-reverse: sinfo -h -N -o "%T" -n $NODE | tr -d '*~#$@+'
+list: sinfo -h -o "%T" | tr -d '*~#!%$@+^-'
+reverse: sinfo -h -N -o "%T" -n $NODE | tr -d '*~#!%$@+^-'
 cache_time: 60
 
 #


conf/groups.conf.d/slurm.conf.example
```diff
@@ -18,8 +18,8 @@ reverse: sinfo -h -N -o "%R" -n $NODE
 [slurmstate,st]
 map: sinfo -h -o "%N" -t $GROUP
 all: sinfo -h -o "%N"
-list: sinfo -h -o "%T" | tr -d '*~#$@+'
-reverse: sinfo -h -N -o "%T" -n $NODE | tr -d '*~#$@+'
+list: sinfo -h -o "%T" | tr -d '*~#!%$@+^-'
```
degremont, line 21: I hope Slurm supports UTF-8, as it will eventually exhaust us-ascii special characters... ;)

Wondering if you should not have a different strategy, where you filter only keeping the standard characters and remove all the other ones?

degremont:
Thanks for the patch. Overall looks good!

Could be nice if you try to keep a similar structure for the commit message with a title which starts with the part of code being touched. See https://github.com/cea-hpc/clustershell/commit/e500d9bea9f72e8fa2f362a08aa35dc0197d989d by example :)

conf/groups.conf.d/slurm.conf.example
kcgthb: I can't wait for emoji-based node states :)

But more seriously, I think the filtering should ideally *not* be based on what Slurm states strings could contain. but rather on what is considered acceptable characters in a group name, namely not those: https://github.com/cea-hpc/clustershell/blob/a1875cb78dc5f579e7053dae9bbcd848cc1c3a1f/lib/ClusterShell/NodeSet.py#L74

But I'm not sure we can get that illegal characters list in a group configuration file, can we?

degremont (approved):
Based on the discussion, i'm ok with the current patch

Issue #442, opened 2020-06-29
Node names get mangled when a node name contains one or more 0 characters


OleHolmNielsen:
We have two nodes listserv01 and listserv3, and the second node name listserv3 gets mangled when taken together with the listserv01 node:

$ clush -w listserv01,listserv3 uname -r
listserv03: 2.6.32-754.30.2.el6.x86_64
listserv01: 4.18.0-193.6.3.el8_2.x86_64

When taken alone there is no issue with the listserv3 node name:

$ clush -w listserv3 uname -r
listserv3: 2.6.32-754.30.2.el6.x86_64

(Note: I had to define a DNS CNAME alias listserv03 pointing to listserv3 as a workaround).

The incorrect leading zero gets added to all subsequent names starting with "listserv":

$ clush -w listserv01,listserv3,listserv2 uname -r
listserv02: ssh: Could not resolve hostname listserv02: Name or service not known
clush: listserv02: exited with exit code 255
listserv03: 2.6.32-754.30.2.el6.x86_64
listserv01: 4.18.0-193.6.3.el8_2.x86_64

Yet another example with "00":

$ clush -w a001,a2 uname -r
a001: 3.10.0-1127.8.2.el7.x86_64
a002: 3.10.0-1127.8.2.el7.x86_64

So it seems that the presence of the "0" character triggers the present bug, where the zeroes get added incorrectly to other node names in the list.

We use this EPEL6 package: clustershell-1.8.3-1.el6_10.noarch 
and this EPEL7 package: clustershell-1.8.3-1.el7.noarch
and this Fedora FC32 package: clustershell-1.8.3-2.fc32.noarch

degremont:
I think your ticket is a duplicate of #293 .

OleHolmNielsen:
Thanks!  This is a pretty surprising bug.

OleHolmNielsen:
I've added references to this issue in my Slurm Wiki page https://wiki.fysik.dtu.dk/niflheim/SLURM#clustershell

thiell:
fixed in https://github.com/cea-hpc/clustershell/commit/5a41bc09f70309600c1a407d2bb3dd08f5d1ba65

PR #473, opened 2022-06-21
RangeSet: support ranges with zero padding of mixed lengths (#293)

### LINKED ISSUE #293: strange behaviour when factorizing (new '0' issue with nodeset)
Hi,

Recently I had some issue with factorization ([#288](https://github.com/cea-hpc/clustershell/issues/288) which was a duplicated of [#286](https://github.com/cea-hpc/clustershell/issues/286)) which I fixed by installing the last version (from master branch). It fixed my issue.

I have now a new one (after eating my '0', it's adding an extra '0'):

```
$ cat liste_failed_for_clush
firstgroupa001
firstgroupa002
secondgroup001
secondgroup002
secondgroup003
secondgroup005
secondgroup006
secondgroup008
namewithissue007
namewithissue008
namewithissue009
namewithissue01
namewithissue02
namewithissue03
namewithissue04
namewithissue10
$ paste -sd, liste_failed_for_clush | nodeset -f
firstgroupa[001-002],namewithissue[001-004,007-010],secondgroup[001-003,005-006,008]
```

It is adding a 0 to the last five servers!
namewithissue[001-004,007-010] should be namewithissue[01-04,10],namewithissue[007-009]

If the servers are not in the same order (starting with 2 digits server's name instead of the 3 digits one):

```
$ cat liste_failed_for_clush2
namewithissue01
namewithissue02
namewithissue03
namewithissue04
namewithissue10
firstgroupa001
firstgroupa002
secondgroup001
secondgroup002
secondgroup003
secondgroup005
secondgroup006
secondgroup008
namewithissue007
namewithissue008
namewithissue009
$ paste -sd, liste_failed_for_clush2 | nodeset -f
firstgroupa[001-002],namewithissue[01-04,07-10],secondgroup[001-003,005-006,008]
```

This time, it's eating a '0'.

As I saw the last version was released (which should be the same I installed 10 days before), I installed it instead, and it is the same behaviour

```
$ rpm -qa | grep clustershell
clustershell-1.7.1-1.el6.noarch
$ /usr/bin/nodeset --version
nodeset 1.7.1
```

I even try to rollback to my initial version (1.7-1, and it is the same).

Is it a new bug?

Thanks

thiell:
This patch adds support for mixed lengths 0-padding ranges by
using sets of strings instead of integers. No need to keep
track of padding length anymore, as this information is self
contained in the strings.

  $ cluset -f bar002,bar01
  bar[01,002]

Deprecated RangeSet.padding which is now a noop.

Closes #293 #442


degremont:
That's look great ! I did not review in details yet, waiting for a more final version, however, my 2 first comments are:
- What about the performance? Worth trying the nodeset/rangeset benchmark we have.
- Be careful about retro-compatibility. We need to be backward compatible with 1.8 IMO. That will make the patch is bit more complex, but we don't really have a choice :/  Additionally, should we raise DeprecationWarnings right away? (btw, we can add a patch to raise DeprecationWarnings for all the places we already have deprecated code already for a while).

doc/sphinx/guide/rangesets.rst
```diff
@@ -57,7 +66,8 @@ tuples, for instance::
     0-5; 10-12
 
     >>> print list(rnd)
-    [(0, 10), (0, 11), (0, 12), (1, 10), (1, 11), (1, 12), (2, 10), (2, 11), (2, 12), (3, 10), (3, 11), (3, 12), (4, 10), (4, 11), (4, 12), (5, 10), (5, 11), (5, 12)]
+    [('0', '10'), ('0', '11'), ('0', '12'), ('1', '10'), ('1', '11'), ('1', '12'), ('2', '10'), ('2', '11'), ('2', '12'), ('3', '10'), ('3', '11'), ('3', '12'), ('4', '10'), ('4', '11'), ('4', '12'), ('5', '10'), ('5', '11'), ('5', '12')]
```
degremont, line 69: (note) non backward compatible :/

degremont, line 69: (note) non backward compatible

lib/ClusterShell/CLI/Nodeset.py
```diff
@@ -287,7 +287,8 @@ def nodeset():
     if options.axis:
         if not options.axis.startswith('-'):
             # axis are 1-indexed in nodeset CLI (0 ignored)
-            xset.fold_axis = tuple(x-1 for x in RangeSet(options.axis) if x > 0)
+            xset.fold_axis = tuple(x-1 for x in \
+                                   RangeSet(options.axis).intiter() if x > 0)
```
degremont, line 291: as `RangeSet().__iter__()` is no more backward compatible, you add to change the CLI code. As other app relying on that will have the same issue :( and upgrading to 1.9 will break it.

lib/ClusterShell/RangeSet.py
```diff
@@ -194,6 +198,14 @@ def fromone(cls, index, pad=0, autostep=None):
             inst.add_range(index.start or 0, index.stop, index.step or 1, pad)
         return inst
 
+    @property
+    def padding(self):
+        warnings.warn("padding has been deprecated", DeprecationWarning)
```
degremont, line 203: i prefer we don't emit warning right now. if you update CS and some user app, like Shine, uses this padding args, there will be warnings everywhere without giving a chance to these apps to be compatible.

lib/ClusterShell/RangeSet.py
```diff
@@ -295,134 +318,128 @@ def __str__(self):
...
-        for i in self._sorted():
-            if k is None:
-                k = j = i
-            if i - j > 1:
-                yield slice(k, j + 1, 1)
-                k = i
-            j = i
-        if k is not None:
-            yield slice(k, j + 1, 1)
+    def slices2(self, autostep=AUTOSTEP_DISABLED):
```
degremont, line 321: to be eventually put in `slices()` ?

lib/ClusterShell/RangeSet.py
```diff
@@ -886,9 +916,7 @@ def _iter(self):
     @precond_fold()
     def iter_padding(self):
         """Iterate through individual items as tuples with padding info."""
-        for vec in self._veclist:
-            for ivec in product(*vec):
-                yield ivec, [rg.padding for rg in vec]
+        warnings.warn("padding is deprecated", DeprecationWarning)
```
degremont, line 919: same here, too early for that

degremont (commented):
I'm worried by the compat breakage. I think this is too risky to release 1.9 with this RangeSet behavior change. (str vs int)
NodeSet is also behaving differently, now that the padding is not use the same way, but hopefully everybody will think this was a bug :)
Worth a beta?

lib/ClusterShell/RangeSet.py
thiell: I will see what I can do to merge `slices()` and `slices2()` :)

lib/ClusterShell/CLI/Nodeset.py
thiell: We'll release an intermediate version 1.8.5 with `intiter()` with a warning to avoid `__iter__()`. That should let some time for clients to adapt.

lib/ClusterShell/RangeSet.py
thiell: Actually, forget about the PendingDeprecationWarning thing. We don't need it yet. I think the padding property might still be useful in some cases, even with mixed length padding... so perhaps this can just stay as part of the API (especially useful to get the max padding length of a set I guess). If needed, we can optimize its performance later.

lib/ClusterShell/RangeSet.py
```diff
@@ -71,25 +71,27 @@ class RangeSet(set):
...
-    instance variable named padding. It may be changed at any time. Padding is
-    a simple display feature per RangeSet object, thus current padding value is
-    not taken into account when computing set operations.
-    RangeSet is itself an iterator over its items as integers (instead of
-    strings). However, this behavior is going to change in the next version
-    v1.9. For compatibility, please use the explicit method
-    :meth:`.RangeSet.intiter` to iterate over the set's indexes as integers.
-    To iterate over string items (with zero-padding if present), please use the
-    :meth:`RangeSet.striter` method.
+    Padding of ranges (eg. "003-009") is inferred from intput arguments and
```
martinetd, line 84: input

lib/ClusterShell/RangeSet.py
```diff
@@ -225,7 +253,8 @@ def dim(self):
 
     def _sorted(self):
         """Get sorted list from inner set."""
-        return sorted(set.__iter__(self))
+        # for mixed padding support, sort by both string length and index
+        return sorted(set.__iter__(self), key=lambda x: (len(x), int(x)))
```
martinetd, line 257: since we compare length first we don't actually need to cast to int here for sorting: "03" > "01" just the same as 3 > 1 for a given length (might save a tiny bit of time)

degremont, line 257: Just wondering if people would be more interesting in simple ordering (`int(x)`) like  `1`, `02`, `2`, `03` ... except if that is needed for folding algorithm?

lib/ClusterShell/RangeSet.py
```diff
@@ -306,134 +344,128 @@ def __str__(self):
...
-        for i in self._sorted():
-            if k is None:
-                k = j = i
-            if i - j > 1:
-                yield slice(k, j + 1, 1)
-                k = i
-            j = i
-        if k is not None:
-            yield slice(k, j + 1, 1)
+    def slices2(self, autostep=AUTOSTEP_DISABLED):
```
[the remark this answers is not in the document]
thiell: Sounds good! Making it private (`_slices_padding()`) for now, we can always see later if there is an interest of having it public.

PR #475, opened 2022-06-28
RangeSet: support ranges with zero padding of mixed lengths (DRAFT btype)

thiell:
DRAFT

RangeSet/RangeSetND with backend type int / str


thiell:
abandoned

PR #478, opened 2022-06-30
clush: always close stdin stream of worker when it is not used

thiell:
Pass explicit stdin boolean flag when calling Task.shell() to fix an
issue where the stdin stream of the worker is never closed, even if
not used.

This change should not impact any already working stdin workflow.

However, the following command from a tty should not block anymore:

$ clush -w nodes cat

### DIFF
--- a/lib/ClusterShell/CLI/Clush.py
+++ b/lib/ClusterShell/CLI/Clush.py
@@ -676,13 +676,13 @@ def run_command(task, cmd, ns, timeout, display, remote, trytree):
         # this is the simpler but faster output handler
         handler = DirectOutputHandler(display)
 
+    stdin = task.default("USER_stdin_worker")
     worker = task.shell(cmd, nodes=ns, handler=handler, timeout=timeout,
-                        remote=remote, tree=trytree)
+                        remote=remote, tree=trytree, stdin=stdin)
     if ns is None:
         worker.set_key('LOCAL')
-    if task.default("USER_stdin_worker"):
+    if stdin:
         bind_stdin(worker, display)
-
     task.resume()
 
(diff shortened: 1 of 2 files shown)


degremont (approved):
Code wise, looks good, but i'm not sure i can say this is a good or bad idea. I'm approving it, but without lots of confidence (what's pdsh behavior here?)

martinetd (approved):
LGTM.

There's just one change of behavior, that a command expecting things from stdin when stdin is not forwarded quits immediately -- I think that's actually good.

```
PYTHONPATH=$PWD/lib/ python3 -m ClusterShell.CLI.Clush -w localhost cat
```
(note even when it wasn't quitting immediately stdin was captured for the interactive side of things, not for cat, so this doesn't break anything)

That aside I couldn't see any difference, and code looks good to me.

Also, I have a deprecation warning on python 3.10
```
/home/martinet/code/clustershell/lib/ClusterShell/CLI/Clush.py:651: DeprecationWarning: setDaemon() is deprecated, set the daemon attribute instead
  stdin_thread.setDaemon(True)
```

thiell:
> Also, I have a deprecation warning on python 3.10

Thanks! I missed that as we don't run the tests with Python 3.10 yet, because python-nose is broken. `setDaemon()` has just been deprecated in 3.10.

The [daemon attribute](https://docs.python.org/2/library/threading.html#threading.Thread.daemon) was added in 2.6, so we can definitely use that now. I will submit a separate patch.

There is also a daemon argument for `threading.Thread()`
Changed in version 3.3: [Added the daemon argument.](https://docs.python.org/3/library/threading.html#threading.Thread) We cannot use that yet as we still support Python < 3.3.

PR #479, opened 2022-07-01
clush: use daemon attribute instead of deprecated setDaemon()

thiell:
setDaemon() has been deprecated since Python 3.10.

Use daemon property of threading.Thread() instead. It is available since
Python 2.6.

Fixes:
    Clush.py:651: DeprecationWarning: setDaemon() is deprecated, set the
    daemon attribute instead


thiell:
Discovered in #478

degremont (approved):

PR #480, opened 2022-07-04
Task: only terminate Task once (#110)

### LINKED ISSUE #110: testThreadTaskWaitWhenNotStarted failure
Investigate and fix the following test. It's working but Python interpreter is hanging at the end (thread not joined?).

``` python
    def testThreadTaskWaitWhenNotStarted(self):
        """test task_wait() when workers not started"""

        for i in range(1, 5):
            task = Task()
            task.shell("sleep %d" % i)

        task_wait()
        task.resume()
```

thiell:
Add a check using existing _terminated attribute to avoid terminating
a Task multiple times, which could cause issues.

Closes #110

### DIFF
--- a/lib/ClusterShell/Task.py
+++ b/lib/ClusterShell/Task.py
@@ -950,7 +950,10 @@ class Task(object):
         """
         Abort completion subroutine.
         """
-        assert self._quit == True
+        assert self._quit is True
+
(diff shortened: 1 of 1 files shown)


thiell:
Superseded by #640

Issue #111, opened 2012-03-03
User configuration files should be compliant with XDG


thiell:
The clustershell user configuration file, should be moved from `~/.clush.conf` to `$XDG_CONFIG_HOME`
To be compliant with the recent specifications.

http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

PR #483, opened 2022-08-03
Introduce $CLUSTERSHELL_CFGDIR 

kcgthb:
Introduce `$CLUSTERSHELL_CFGDIR` as an alternate location for configuration files.

This is basically a revisit of #191. This original issue led to #111, which allows *per-user* configuration file overrides, but doesn't provide a way to define system-wide configuration files (not per-user), which are not stored in `/etc`.

The use case is a multi-node cluster, where ClusterShell could be provided as a user-facing tool, installed on a shared file system (like a NFS mount point), and not as a local package on each node. In that case, providing system defaults for all users would requiring distributing configuration files on each and every of the cluster node's local `/etc/` directory. A more flexible approach would be to define an environment variable (the proposed `$CLUSTERSHELL_CFGDIR`) to point to the global configuration directory, which could be co-located with the shared ClusterShell installation.

When `$CLUSTERSHELL_CFGDIR` is not defined, the default `/etc/clustershell` is used as a fallback to keep compatibility with existing configurations.

This PR:
* updates `lib/ClusterShell/Defaults.py`
* updates (text) documentation files in `doc/`
* adds a test in `tests/CLIConfigTest.py`

### DIFF
--- a/doc/examples/defaults.conf-rsh
+++ b/doc/examples/defaults.conf-rsh
@@ -4,7 +4,8 @@
 # Example defaults.conf file for clusters using rsh instead of ssh.
 #
 # To enable this file, install it in one of the following locations:
-#       /etc/clustershell/defaults.conf (system-wide)
+#       $CLUSTERSHELL_CFGDIR/defaults.conf (global configuration, default to
+#       /etc/clustershell/defaults.conf)
 #       $XDG_CONFIG_HOME/clustershell/defaults.conf (per-user)
 #       $HOME/.local/etc/clustershell/defaults.conf (per-user)
 #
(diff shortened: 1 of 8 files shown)


degremont (approved):
Just ensure the new variable name matches other possible variables we already support, if they exist. (CLUSTERSHELL_ prefix).

kcgthb:
> Just ensure the new variable name matches other possible variables we already support, if they exist. (CLUSTERSHELL_ prefix).

Yes. The only other environment variables I could find are the `CLUSTERSHELL_GW_*` env vars, so we should be god.
https://github.com/cea-hpc/clustershell/blob/master/lib/ClusterShell/Worker/Tree.py#L179-L182

thiell (approved):

PR #484, opened 2022-08-06
Fix for python-3.10

bjornfor:
...by using current_thread() and notify_all() instead of the deprecated
currentThread() and notifyAll() functions. This fixes
DeprecationWarnings from the test suite that end up as errors.

AFAICT, the new names have been available since python-2.6[1], so this
change should not break support for any current python versions.

[1] https://docs.python.org/2.6/library/threading.html#threading.current_thread
    https://docs.python.org/2.6/library/threading.html#threading.Condition.notify_all

### DIFF
--- a/lib/ClusterShell/Task.py
+++ b/lib/ClusterShell/Task.py
@@ -258,7 +258,7 @@ class Task(object):
             self._cond.acquire()
             try:
                 self.suspend_count = min(self.suspend_count, 0)
-                self._cond.notifyAll()
+                self._cond.notify_all()
             finally:
                 self._cond.release()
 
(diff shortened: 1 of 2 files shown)


bjornfor:
The latest https://github.com/nixos/nixpkgs/ defaults to python-3.10 (which means clustershell is currently broken), so landing this PR and making a release would be appreciated :-)

thiell (approved):

thiell:
@bjornfor thanks much! it's good timing as we'll release version 1.9 soon. It will include these fixes for Python 3.10.

bjornfor:
Thanks!

PR #487, opened 2022-09-08
CLIConfigTest: Use integer division for soft limit computation

arkamar:
The `CLI.Config.ClushConfig (setrlimit)` test fails with `python3.10` with an error:

```
       resource.setrlimit(resource.RLIMIT_NOFILE, (hard2/2, hard))
   TypeError: 'float' object cannot be interpreted as an integer
```
The integer division solves the problem.


thiell (approved):
Thanks!

PR #490, opened 2022-11-06
Disable python2 on SLES15sp4 and newer

mattaezell:
py2 has been removed from SLES as of 15sp4
https://www.suse.com/releasenotes/x86_64/SUSE-SLES/15-SP4/index.html#jsc-SLE-16747


thiell:
thanks @mattaezell

PR #492, opened 2022-11-20
clush: add support for run modes (clush --mode)

### LINKED ISSUE #198: clush: support password-based authentication
support password-based authentication for initial connection to nodes

### LINKED ISSUE #234: clush: Interactive usage - sudo with password and requiretty
Hi,

I have used [polysh](http://guichaz.free.fr/polysh/) for several years while maintaining a small data center with CentOS servers. Today I switched to clustershell/clush, since I just started maintaining a tiny data center with Debian servers and polysh isn't packaged by Debian. clush seems to be a very good replacement, but I struggle with interactive usage and sudo with password when requiretty is enabled.

If I use

```
clush -o '-t -t' -w node[1-10]
```

the normal "sudo: no tty present and no askpass program specified" error is gone when I run "sudo date", but the password isn't captured if I enter it (and the password is displayed).

If I on the same setup use polysh

```
./polysh.py 'node<1-10>'
```

there is no problem with entering the password, and there is even a command ":hide_password" that stops the password from displaying.

Is there some way clush can handle this situation? If not, consider this a feature request. Maybe the code in polysh can be reused?

Added: Polysh 0.4, Clustershell 1.6

### LINKED ISSUE #423: Permission denied (publickey,password) problem
as root, the first of the following works and the second no:
1) clush -w node-10 uptime
node-10:  12:12:45 up 21 days, 23:27,  0 users,  load average: 21.04, 21.01, 21.00
2) clush --user someuser -w node-10 uptime
node-10: Permission denied (publickey,password).
clush: node-10: exited with exit code 255

/etc/clustershell/clush.conf is the default one

thiell:
Add support for configuration-based run modes for clush. Run modes are disabled by default because they rely on external tools that we don't want to depend on. They can easily be enabled if needed by installing the corresponding .conf files in clush.conf.d (clush.conf's confdir).

sudo support is provided via an example run mode, by defining two variables command_prefix and password_prompt:

  [mode:sudo]
  password_prompt: yes
  command_prefix: /usr/bin/sudo -S -p "''"

Then, the sudo mode is activated with --mode or -m:

  $ clush -w nodes -m sudo command

Also added an example of run mode for password-auth based auth with sshpass.

Main changes:

  clush.conf:
  - add confdir to define paths to config files for run modes
  - add command_prefix
  - add password_prompt
  - removed sudo_command

  clush.conf.d:
  - add sshpass.conf.example
  - add sudo.conf.example

  clush:
  - added --mode
  - removed --sudo

Closes #198.
Closes #423.
Fixes #234.


thiell:
for the online doc, see preview at:
https://clustershell-test.readthedocs.io/en/latest/config.html#clush
and
https://clustershell-test.readthedocs.io/en/latest/tools/clush.html#run-modes

PR #514, opened 2022-12-09
Fix typos found with codespell

xdelaruelle:
Fix typos found with the codespell utility [1].

[1] https://github.com/codespell-project/codespell

### DIFF
--- a/doc/sphinx/config.rst
+++ b/doc/sphinx/config.rst
@@ -48,7 +48,7 @@ The following table describes available ``clush`` config file settings.
 |                 | :ref:`run modes <clushmode-config>` that can then  |
 |                 | be activated with `--mode`. All other ``clush``    |
 |                 | config file settings defined in this table might   |
-|                 | be overriden in a run mode. Each mode section      |
+|                 | be overridden in a run mode. Each mode section     |
 |                 | should have a name prefixed by "mode:" to clearly  |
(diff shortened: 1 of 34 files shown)


doc/sphinx/config.rst
```diff
@@ -48,7 +48,7 @@ The following table describes available ``clush`` config file settings.
 |                 | :ref:`run modes <clushmode-config>` that can then  |
 |                 | be activated with `--mode`. All other ``clush``    |
 |                 | config file settings defined in this table might   |
-|                 | be overriden in a run mode. Each mode section      |
+|                 | be overridden in a run mode. Each mode section      |
```
thiell, line 51: unaligned table might break sphinx

tests/TaskLocalMixin.py
```diff
@@ -568,7 +568,7 @@ def testTaskAbortSelf(self):
 
         # retry
         task = task_self()
-        worker = task.shell("/bin/echo shouldnt see that")
+        worker = task.shell("/bin/echo shouldn't see that")
```
thiell, line 571: we should just avoid the simple quote here, to make things safer

xdelaruelle: ok, I propose to change *shouldn't* into *should not* in revised commit.

thiell:
Prefect, thanks @xdelaruelle!

PR #523, opened 2023-04-19
Use python3 if python is not installed

mattaezell:
SLES15sp4 doesn't have `python` installed, so fall back to `python3`


thiell (approved):
Thanks!

PR #536, opened 2023-09-19
Topology: check that node groups/wildcards are non-empty (#527)

### LINKED ISSUE #527: "topology.conf" leads to high CPU usage and memory not being released
My 'topology.conf' config file has ? or * in gateway names like below:

[routes]
Controller-vm1:Controller-vm2,Controller-vm3
Controller-vm2**?**:Computer101

After executing command: "env/bin/clush -d -w 127.1 whoami", an error occurred and the command is stuck, which leads to high CPU usage and memory not being released. The debug output is below:
```
Changing max open files soft limit from 65535 to 8192
User interaction: True
Create STDIN worker: False
DEBUG:ClusterShell.NodeUtils:EXEC '[ -f /etc/clustershell/groups ] && f=/etc/clustershell/groups || f=/etc/clustershell/groups.d/local.cfg; sed -n 's/^all:\(.*\)/\1/p' $f'
DEBUG:ClusterShell.NodeUtils:READ ''
```

Please help me check this issue, thank you very much.

thiell:
Router and destination node sets defined in topology.conf may use NodeSet groups and wildcards but those have to resolve to non-empty node sets when we build the tree.

Add checks for this in the code and add a note to the documentation.

Fixes #527.

### DIFF
--- a/lib/ClusterShell/Topology.py
+++ b/lib/ClusterShell/Topology.py
@@ -42,7 +42,12 @@ except ImportError:
     # Python 2 compat
     import ConfigParser as configparser
 
+import logging
+
 from ClusterShell.NodeSet import NodeSet
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 class TopologyError(Exception):
@@ -105,6 +110,7 @@ class TopologyNodeGroup(object):
         its parent
         """
         assert isinstance(child, TopologyNodeGroup)
+        assert len(child.nodeset) > 0, "empty nodeset in child"
 
         if child in self._children:
             return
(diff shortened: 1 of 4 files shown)


lib/ClusterShell/Topology.py
```diff
@@ -456,7 +456,17 @@ def _build_graph(self):
         """
         self.graph = TopologyGraph()
         for src, dst in self._topology:
-            self.graph.add_route(NodeSet(src), NodeSet(dst))
+            # GH#527: router and destination node sets may use NodeSet groups
+            # but they have to resolve to something when we build the graph!
```
degremont, line 460: I'm wondering if this is really a big deal. Couldn't we just ignore those routes? If one of src or dst is empty, just don't `add_route()` ?

thiell: Great point. It might be more flexible to just ignore such route, I will change that, thanks!

degremont (approved):

PR #555, opened 2024-03-03
Drop support for insecure end-of-life Python <=3.7

hartwork:
Hi!

I'm not sure how welcome this is, I just became aware that clustershell tries supporting some insecure end-of-life versions of Python, so I decided to offer this pull request to get rid of that baggage, and see what you think :beers: 

Best, Sebastian

### DIFF
--- a/.github/workflows/nosetests.yml
+++ b/.github/workflows/nosetests.yml
@@ -8,10 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11']
-        include:
-          - os: ubuntu-20.04
-            python-version: '3.6'
+        python-version: ['3.8', '3.9', '3.10', '3.11']
     name: Tests with Python ${{ matrix.python-version }}
     steps:
     - uses: actions/checkout@v3
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Requirements
 
  * GNU/Linux, BSD, Mac OS X
  * OpenSSH (ssh/scp) or rsh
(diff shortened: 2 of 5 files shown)


degremont:
Hello

Thanks for your proposal but I don't think that's the right way to drop support. Except for the Display.py change, all the code is still fully functional under Python 2.7 and 3.6+, you're just removing the doc which says we support those changes, but the code base is not touched, so the code still work there.

I don't see any benefit of dropping this support is the code is still working fine there, for people who needs that.
There is still a large RHEL7 install base that needs Python 2 support.

What would be great when that support will be dropped is to get rid of all the compatibility code we have for Py2 and Py3 <= 3.7. 

The plan is to drop that support for a next minor release (1.10 ?) . There we would drop Py2, and remove all the code we have which is there only to support Py2. That would be beneficial to us as the code to maintain will be simpler. We would prefer to see that kind of patch, dropping py2 support first. We have normal "py2" tag in the source code for that, but a real code audit would be useful to catch the places where py2 is not explicit.

hartwork:
Hi @degremont, thanks for your reply! I do not agree with your assessment — I don't consider systems without security updates worthy of support by anyone — but I accept your choice, and will close this pull request.

thiell:
@hartwork We really appreciate that you contacted us regarding this and thanks for the PR.

Some additional context: older versions of Python are still supported by some Linux distributions, like in RHEL and spinoffs, and critical security updates are available. Because clustershell is mainly used as a system-level tool, installed system-wide as distribution packages, we're a bit conservative when it's time to drop support for Python versions, and we prefer to not break support for older Python versions unless it is really needed. But I agree with you that it is time to make some changes in ClusterShell to focus on the newer versions of Python, and we plan to work on that soon.

PR #556, opened 2024-03-03
Fix tests for libexpat >=2.6.0 (or with CVE-2023-52425 fixes backported)

hartwork:
Hi! :wave: 

It was reported that the clustershell test suite breaks with Expat >=2.6.0 (https://bugs.gentoo.org/924601). So this pull request fixes the test suite for Expat >=2.6.0 with CPython 3.13 and backport releases by using the new pyexpat API (https://github.com/python/cpython/issues/115398). There is no rush in merging this, we can definitey wait for that APi to become more widely available if you like, it will likely appear in these very CPython releases:

- 3.8.19
- 3.9.19
- 3.10.14
- 3.11.9
- 3.12.3
- 3.13.0

I should note that Python 3.7 and earlier are dead, hence related pull request #555.
I'm looking forward to your review, maybe I missed something :beers: 

Best, Sebastian

### DIFF
--- a/tests/TreeGatewayTest.py
+++ b/tests/TreeGatewayTest.py
@@ -135,6 +135,8 @@ class TreeGatewayBaseTest(unittest.TestCase):
                 break
             self.assertTrue(type(xml_msg) is bytes)
             self.parser.feed(xml_msg)
+            if hasattr(self.parser, 'flush'):  # >=3.13 and backports
+                self.parser.flush()
 
         return self.xml_reader.pop_msg()
 
--- a/tests/TreeGatewayTest.py
+++ b/tests/TreeGatewayTest.py
@@ -131,7 +131,8 @@ class TreeGatewayBaseTest(unittest.TestCase):
         while not self.xml_reader.msg_available():
             xml_msg = self.gateway.recv()
             if len(xml_msg) == 0:
-                return None
+                self.parser.close()
+                break
             self.assertTrue(type(xml_msg) is bytes)
             self.parser.feed(xml_msg)
 


hartwork:
Quick update on CPython: All related backport pull requests have now been merged. The next step is CPython security releases and then distros picking them up.

Regarding related expected releases dates, so far I only found:
- 3.11.9: Monday, 2024-**04-01** — https://peps.python.org/pep-0664/
- 3.12.3: Tuesday, 2024-**04-09** — https://peps.python.org/pep-0693/
- 3.13.0 final: Tuesday, 2024-**10-01** — https://peps.python.org/pep-0719/

hartwork:
@thiell by any chance do you know if the issue I'm fixing for the tests here affects the core ClusterShell software also? Is it just the tests that break with Expat >=2.6.0 or more?

hartwork:
> * 3.8.19
> * 3.9.19
> * 3.10.14

PS: These are out since yesterday March 19, 2024:
- 3.8.19 — https://www.python.org/downloads/release/python-3819/
- 3.8.19 — https://www.python.org/downloads/release/python-3919/
- 3.10.14 — https://www.python.org/downloads/release/python-31014/

hartwork:
> @thiell by any chance do you know if the issue I'm fixing for the tests here affects the core ClusterShell software also? Is it just the tests that break with Expat >=2.6.0 or more?

@thiell any chance for a reply this week? We'll have to make decision in Gentoo whether to stabilize libexpat >=2.6.0 or not for all users, and knowing if core ClusterShell is affected outside the tests is the last missing piece in the picture, before making a decision. Thanks in advance, have a good start to the week :beers:

thiell:
@hartwork When ClusterShell is configured in [tree execution mode](https://clustershell.readthedocs.io/en/latest/tools/clush.html#tree-execution-mode), the communication protocol used with the remote ClusterShell gateway(s) uses XML and we parse it incrementally with use an `xml.sax.xmlreader.IncrementalParser`. So the only instance of `IncrementalParser.feed()` in the core ClusterShell code is done here: https://github.com/cea-hpc/clustershell/blob/master/lib/ClusterShell/Communication.py#L231

A question I have is does the issue only occur with multithreaded codes? The test case uses two threads for example to simulate a remote gateway. In `Communication.py`, unlike the test in `tests/TreeGatewayTest.py`, a single thread is used and the producer of the XML content is remote, on a different machine.

hartwork:
@thiell thanks for your detailed reply! :+1: 

The way I read the linked code, `self._xml_reader` is only used as an event FIFO that is emptied at the speed of availability at: https://github.com/cea-hpc/clustershell/blob/ec8be763791ceb1c546827ad9e3b37ab93b1ed45/lib/ClusterShell/Communication.py#L254-L258

To me, it looks like the parser can just have its pace and the receivers need to be okay with any rate. If that's reality, it probably means that there is no problem with an occasional delay in events?

With regard to multi-threading, my understanding is that it's technically not the two threads but their blocking lockstep relationship that breaks with Expat >=2.6.0: every parser feed was expected to produce events immediately by the test code. The fix to CVE-2023-52425 in Expat can add delay to creation of events — firing of handlers in the language of Expat — depending on how big or small input chunks get, and that made one thread wait for an event that would not be produced without further input.

Maybe a good stress test would be to turn…

https://github.com/cea-hpc/clustershell/blob/ec8be763791ceb1c546827ad9e3b37ab93b1ed45/lib/ClusterShell/Communication.py#L231

…into a loop feeding single bytes to make things worst (and be reverted), and running that with Expat >=2.6.0, if someone gets a chance to.

What do you think?

PS: On a side note Expat is not thread-safe and multi-threaded access would need synchronization around it, likely even from Python code (https://github.com/python/cpython/issues/62170), not just in C.

hartwork:
> * 3.11.9: Monday, 2024-**04-01** — https://peps.python.org/pep-0664/

Released by now on 2024-04-02 — https://docs.python.org/release/3.11.9/whatsnew/changelog.html#python-3-11-9

> * 3.12.3: Tuesday, 2024-**04-09** — https://peps.python.org/pep-0693/

Released by now on 2024-04-09 — https://docs.python.org/release/3.12.3/whatsnew/changelog.html#python-3-12-2

hartwork:
@thiell I would expect CI on master to fail by now because it has `2.4.7-1ubuntu0.3` installed for all but Python 3.6 and that comes with the CVE-2023-52425 fixes according to https://changelogs.ubuntu.com/changelogs/pool/main/e/expat/expat_2.4.7-1ubuntu0.3/changelog but for some reason it's no failing — any ideas why? And: How do you feel about this pull request today?

thiell:
> PS: On a side note Expat is not thread-safe and multi-threaded access would need synchronization around it, likely even from Python code ([python/cpython#62170](https://github.com/python/cpython/issues/62170)), not just in C. (I'm in favor of considering that a bug in CPython.)

Thanks for the clarification. That's fine in our case as we communicate through pipes.


> @thiell I would expect CI on master to fail by now because it has `2.4.7-1ubuntu0.3` installed for all but Python 3.6 and that comes with the [CVE-2023-52425](https://github.com/advisories/GHSA-gh68-jm46-84rf) fixes according to https://changelogs.ubuntu.com/changelogs/pool/main/e/expat/expat_2.4.7-1ubuntu0.3/changelog but for some reason it's no failing — any ideas why? And: How do you feel about this pull request today?

No idea why... Our own tests were not failing before but we probably did not test with expat >=2.6. I just added support for testing with Python 3.12 and I am seeing occasional hung tests, so this might be related.

Otherwise I reviewed your proposed changes and I am ok to merge them as a first step. Note that I personally don't love the additional `flush()` method which is called after each `feed()` + need to be checked with `hasattr()`. I prefer the `Get/SetReparseDeferralEnabled()` API, which can restore the previous behavior.

Now apart from this test, I think we could have some problems with delayed events in the parser when used in clustershell's tree mode: XML is parsed incrementally between cluster nodes and the connection might be established for a long time between then, triggering events when they arrive. The source XML is generated by clustershell itself, so I think it would be fine in terms of security to just disable this reparse deferral feature with [SetReparseDeferralEnabled(false)](https://docs.python.org/release/3.11.9/library/pyexpat.html#xml.parsers.expat.xmlparser.SetReparseDeferralEnabled) if available.

Sorry to be a bit slow but thanks for your help with this!

thiell:
@hartwork Ah, interestingly, it looks like the incremental SAX parser that we use (`ExpatParser`) didn't get the same API than the `xmlparser` (xml.parsers.expat.xmlparser). With Python 3.12.5, I get the following error when trying to call `ExpatParser.SetReparseDeferralEnabled()`:
```
  File "/home/runner/work/clustershell/clustershell/lib/ClusterShell/Communication.py", line 194, in __init__
    self._parser.SetReparseDeferralEnabled(False)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ExpatParser' object has no attribute 'SetReparseDeferralEnabled'
```

hartwork:
> @hartwork Ah, interestingly, it looks like the incremental SAX parser that we use (`ExpatParser`) didn't get the same API than the `xmlparser` (xml.parsers.expat.xmlparser).

@thiell correct, the description of https://github.com/python/cpython/pull/115623 has the list of all related methods added (and to which class). The involved classes are quite different and there was not much room for how to implement things in CPython.

> With Python 3.12.5, I get the following error when trying to call `ExpatParser.SetReparseDeferralEnabled()`:
> 
> ```
>   File "/home/runner/work/clustershell/clustershell/lib/ClusterShell/Communication.py", line 194, in __init__
>     self._parser.SetReparseDeferralEnabled(False)
>     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'ExpatParser' object has no attribute 'SetReparseDeferralEnabled'
> ```

Correct, there is no such method in that class.

Could you summarize status and plans for the new pull request #569? That would be great!

thiell:
*Note: this comment was generated by Anthropic's Claude (Fable 5), which
also assisted the investigation below — running the elimination matrix
across EL9/EL10 interpreters, writing the expat probe and building the
deterministic reproducer. Posted after my review.*

Following up on this thread, which I left with a few loose ends — including
a question from @hartwork that I never really answered. I recently
investigated intermittent TreeGatewayTest hangs on EL9 (RHEL/Rocky 9);
everything traced back to the reparse deferral discussed here, so
documenting the findings in this thread.

Key points:

- **@hartwork asked in March 2024 whether core ClusterShell is affected
  beyond the tests, and I never gave a definitive answer. Now I can: core
  is NOT affected — only the test suite is.** `Channel.ev_read()` feeds
  complete, newline-framed lines to the parser (workers use line-based
  readers), so tokens never span `feed()` calls and deferral never gets a
  chance to wedge — even where `flush()` is broken (see below). The
  defensive `flush()` added by #569 stays useful where it works.
- **The single-byte stress test @hartwork suggested in this thread: we
  effectively ran it.** Feeding the test's parser one byte at a time makes
  the hang fully deterministic.
- **The "random hung tests" I mentioned in September 2024 are now
  explained** — same mechanism, hitting CI before this PR merged.
- **New distribution twist: on EL9 the fix from this PR cannot work.**
  RHEL 9's expat 2.5.0-2+ backports the CVE-2023-52425 mitigation ("Fix
  parsing of large tokens", RHEL-29699), so the >= 2.6.0 deferral behavior
  is present despite the 2.5.0 version string — but
  `XML_SetReparseDeferralEnabled` is exported and appears to be a no-op,
  which makes CPython's `ExpatParser.flush()` (gh-115623, the API this PR
  relies on) silently ineffective on EL9 with any Python version
  (reproduced with RHEL-built 3.9.25, 3.11.13 and 3.12.13).

Minimal probe for the EL9 behavior (works everywhere except EL9, where both
prints stay empty):

```python
import xml.sax, pyexpat

class H(xml.sax.handler.ContentHandler):
    def __init__(self): self.seen = []
    def startElement(self, name, attrs): self.seen.append(name)

chunks = [b"<?xml version='1.0' encoding='utf-8'?>\n",
          b"<channel", b" version='1.10'", b">"]
h = H()
p = xml.sax.make_parser(["IncrementalParser"])
p.setContentHandler(h)
for c in chunks:
    p.feed(c)
p.flush()
print("after flush():", h.seen)   # EL9: []   elsewhere: ['channel']

seen = []
rp = pyexpat.ParserCreate()
rp.StartElementHandler = lambda n, a: seen.append(n)
rp.SetReparseDeferralEnabled(False)
for c in chunks:
    rp.Parse(c, False)
print("deferral disabled:", seen)  # EL9: []   elsewhere: ['channel']
```

Results: EL9 expat 2.5.0-5 + py3.9/3.11/3.12 → `[]` both times;
EL10 expat 2.7.3 (py3.12) and Debian 13 expat 2.7.1 (py3.13) → `['channel']`.

Mechanism of the test hang, for the record: TreeGatewayTest feeds raw
`os.read()` chunks to its IncrementalParser. When a read lands between the
gateway's chunked writes of the `<channel ...>` start tag, the split token
is deferred and — the protocol being request-reply — no further bytes ever
arrive, so the test hangs. On EL9 `flush()` cannot rescue it, which is why
the hang is intermittent (a scheduling race per channel open) and EL9-only.

We are keeping the test's raw-chunk feeding as-is since it usefully
surfaces the platform issue; PR #664 adds a recv timeout so it fails fast
with a traceback instead of hanging the suite.

PR #561, opened 2024-05-15
Additional Slurm binding for mapping nodes to reservations

kcgthb:
To list nodes in reservation `foobar`
```
$ cluset -f @sr:foobar
sh02-01n60
```

To list reservations that node `sh03-01n72` is part of:
```
$ cluset -ll -s slurmresv sh03-01n72
@slurmresv:test sh03-01n72
$ cluset -f @slurmresv:test
sh03-01n[71-72]
```

### DIFF
--- a/conf/groups.conf.d/slurm.conf.example
+++ b/conf/groups.conf.d/slurm.conf.example
@@ -19,7 +19,6 @@ reverse: sinfo -h -N -o "%R" -n $NODE
 map: scontrol -o show reservation $GROUP | grep -Po 'Nodes=\K[^ ]+'
 all: scontrol -o show reservation | grep -Po 'Nodes=\K[^ ]+'
 list: scontrol -o show reservation | grep -Po 'ReservationName=\K[^ ]+'
-reverse: scontrol -o show node $NODE | grep -Po 'ReservationName=\K[^ ]+' 2>/dev/null || true
 cache_time: 60
 
 #
--- a/conf/groups.conf.d/slurm.conf.example
+++ b/conf/groups.conf.d/slurm.conf.example
@@ -16,8 +16,8 @@ reverse: sinfo -h -N -o "%R" -n $NODE
 # SLURM node-reservation bindings
 #
 [slurmresv,sr]
-map: scontrol -o show reservation $GROUP | grep -Po 'Nodes=\K[^ ]+' | xargs | tr ' ' ','
-all: scontrol -o show reservation | grep -Po 'Nodes=\K[^ ]+' | tr '\n' ','
+map: scontrol -o show reservation $GROUP | grep -Po 'Nodes=\K[^ ]+'
+all: scontrol -o show reservation | grep -Po 'Nodes=\K[^ ]+'
 list: scontrol -o show reservation | grep -Po 'ReservationName=\K[^ ]+'
 reverse: scontrol -o show node $NODE | grep -Po 'ReservationName=\K[^ ]+' 2>/dev/null || true
 cache_time: 60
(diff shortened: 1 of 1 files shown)


conf/groups.conf.d/slurm.conf.example
```diff
@@ -12,6 +12,16 @@ all: sinfo -h -o "%N"
 list: sinfo -h -o "%R"
 reverse: sinfo -h -N -o "%R" -n $NODE
 
+#
+# SLURM node-reservation bindings
+#
+[slurmresv,sr]
+map: scontrol -o show reservation $GROUP | grep -Po 'Nodes=\K[^ ]+' | xargs | tr ' ' ','
```
degremont, line 19: why `xargs | tr` ? This should only return 1 nodeset already?

degremont (commented):
Using json and `jq` would have been another option, but this is probably not widely installed compared to `grep`.

kcgthb:
> Using json and `jq` would have been another option, but this is probably not widely installed compared to `grep`.

I thought about using the JSON output too, but the dependency on `jq` would be a bit much, indeed, as it's far less commonly installed than `grep`.

conf/groups.conf.d/slurm.conf.example
kcgthb: That's an excellent question! :)
I think I was trying to aggregate the returned nodeset on a single line in case of multi-line output (scripting habit) but you're right that it's not necessary here, since that command should return a single nodeset already. And I don't think it's necessary in the `all:` binding either. 
Fix incoming, thanks for catching that!

conf/groups.conf.d/slurm.conf.example
```diff
@@ -12,6 +12,16 @@ all: sinfo -h -o "%N"
...
 reverse: sinfo -h -N -o "%R" -n $NODE
 
+#
+# SLURM node-reservation bindings
+#
+[slurmresv,sr]
+map: scontrol -o show reservation $GROUP | grep -Po 'Nodes=\K[^ ]+'
+all: scontrol -o show reservation | grep -Po 'Nodes=\K[^ ]+'
+list: scontrol -o show reservation | grep -Po 'ReservationName=\K[^ ]+'
+reverse: scontrol -o show node $NODE | grep -Po 'ReservationName=\K[^ ]+' 2>/dev/null || true
```
degremont, line 22: unfortunately, it seems Slurm only shows 1 resa per node. If you let ClusterShell do the reverse automatically, it will handle the case where a node is in several resa gracefully.

kcgthb: Ah that's true: I didn't think about overlapping reservations (having the same node part of multiple simultaneous reservations is probably not that common), but `scontrol show node` only seems to return one reservation in that case, indeed. I guess that's a bug? 
I reported https://bugs.schedmd.com/show_bug.cgi?id=19884, let's see what SchedMD has to say,

degremont: Mhhh.... the field is named `ReservationName`, singular. I don't expect it to return more than 1 name. And even if they confirmed this is a bug, we don't want to wait for the fix to be broadly available. Better merging this PR without `reverse` and maybe change it in the future.

kcgthb: > the field is named ReservationName, singular. I don't expect it to return more than 1 name. 

Aah but that's not the only one, far from it :smile: 
```
$ scontrol show job 46818246 | grep -Po Partition=[a-z,]+
Partition=normal,hns
```

> And even if they confirmed this is a bug, we don't want to wait for the fix to be broadly available. Better merging this PR without reverse and maybe change it in the future.

I agree: the bug is confirmed, but it will most likely not be fixed very soon, so I'll remove the `reverse` command for now. Although I think it would still be more efficient to use it, as I understand it, to minimize calls to Slurm commands (one call to `scontrol` with an explicit `reverse`, vs. two to compute the result from `map` and `list`).

PR #563, opened 2024-06-30
bash completions

martinetd:
Just clush for now; might be some bugs but seems to work.

perf-wise it calls `cluset -L` on every tab, we might need to cache the result somewhere if it is slow in some setups.
(But if user changes conf files it won't be reloaded so that's a bit annoying, it's better not to cache if it's reasonable e.g. just reading the conf files)

Prereq patch makes `cluset -s foo -L` not list `foo` groups with the `@foo:` prefix, like default source with `-L` -- this allows `clush -s foo -w @bar` to complete properly.

### DIFF
--- a/bash_completion.d/cluset
+++ b/bash_completion.d/cluset
@@ -1,0 +1,79 @@
+# cluset bash completion
+#
+# to install in /usr/share/bash-completion/completions/ or ~/.local/share/bash-completion/completions/
+_cluset()
+{
+	# shellcheck disable=SC2034 # set/used by _init_completion
+	local cur prev words cword split
+	local word options="" skip=argv0 groupsource="" cleangroup=""
+
+	_init_completion -s -n : || return
+
+	# stop parsing if there had been any non-option before (or --)
+	for word in "${words[@]}"; do
+		case "$skip" in
+		"") ;;
+		groupsource)
+			groupsource="$word"
+			;& # fallthrough
+		*)
+			skip=""
+			continue
+			;;
+		esac
+		case "$word" in
+		"") ;;
+		--) return;;
+		# no-arg options
+		--version|-h|--help|-n|--nostdin|-a|--all|-q|--quiet|\
+		-v|--verbose|-d|--debug) ;;
+		# get source separately...
+		--groupsource=*) groupsource="${word#*=}";;
+		-s|--groupsource) skip=groupsource;;
+		# assume all the rest as options...
+		# options with = included in word
+		--*=*) ;;
+		-*) skip=any;;
+		*) return;; # was non-option
+		esac
+	done
+
+	case "$prev" in
+	-c|--count|-e|--expand|-f|--fold|\
+	-x|--exclude|-i|--intersection|-X|--xor)
+		case "$cur" in
+		*:*)
+			groupsource="${cur%%:*}"
+			groupsource="${groupsource#@}"
+			;;
+		*)
+			if [ -n "$groupsource" ]; then
+				cleangroup=1
+			fi
+			;;
+		esac
+		options="$(cluset ${groupsource:+-s "$groupsource"} --completion @*)"
+		if [ -n "$cleangroup" ]; then
+			options=${options//@"$groupsource":/@}
+		fi
+		;;
+	-s|--groupsource)
+		options=$(cluset --groupsources --quiet)
+		;;
+	# no-arg options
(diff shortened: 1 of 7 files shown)


bash_completion.d/clush
```diff
@@ -0,0 +1,71 @@
...
+		-*) skip=any;;
+		*) return;; # was non-option
+		esac
+	done
+
+	case "$prev" in
+	-w|-x)
+		options="$(cluset ${groupsource:+-s "$groupsource"} -L 2>/dev/null)"
+		;;
+	_g|--group|-X)
```
degremont, line 45: (typo) `_` -> `-`

thiell:
Hi @martinetd and thanks for this!

I would like to include clush bash completion is 1.9.3 if possible, so I've been looking at this today.
I confirm it works however there is one problem I noticed when testing on our cluster Sherlock: `-L` is way too costly. On our main production cluster, a sysadmin you know defined 32 group sources :-), including various external group sources. Calling `clush -L` will then try to resolve all groups in all sources and takes up to several minutes to complete.

But I think we might be able to come up with something to only call `cluset -l` when the user presses TAB (and never `-L`). But I'm not 100% sure this is doable technically and need your help. Let me explain below.

The idea would be as follow:

* `clush -s source -w <TAB>`
Similar to what you propose but I  just added `-G` so that the group source is not repeated (open for discussion):
```
options="$(cluset -s "$groupsource" -l -G 2>/dev/null)"  
```
This gives us this on our cluster:
```
# clush -s pdu -w <TAB>
```
```
# clush -s pdu -w @PX
@PX2-1901U-N1A6  @PX3-1901U-N1    @PX3-1901U-N1A6  @PX4-5851-E7V2 
```

* Now, for `-w` when `-s source` is not specified, the idea would be to return groups in the default group sources + "@source:" prefixes for the matching group sources. Something like that:

```
# clush -w <TAB>
@base           @cpu            @hw:            @island:    ....
```

Here `@base` and `@cpu` are normal groups in the default group source, and `@hw:` and `@island:` are group sources. 

I got this result from this:

```
    -w|-x)                                                                      
        if [[ -n "$groupsource" ]]; then                                        
            options="$(cluset -s "$groupsource" -l -G 2>/dev/null)"             
        else                                                                    
            options="$(cluset --groupsources | sed -e 's/ (default)//' -e 's/^/@/' -e 's/$/:/')"
            options+="$(cluset -l 2>/dev/null)"                                 
        fi                                                                      
        ;;  
```

However, I'm not sure how to trigger a group lookup when we do `clush -w @hw:<TAB>`. First, bash completion adds a space if I hit TAB after `@hw:` (that's the first problem) and second, we likely need to call `cluset -s hw -l -G` to list all groups in the `hw` source in that case. I'm not sure how to do that. If you have any ideas on how to resolve this... let me know. Should we define another completion function and use `complete` again?

bash_completion.d/cluset
```diff
@@ -0,0 +1,79 @@
...
+			fi
+			;;
+		esac
+		options="$(cluset ${groupsource:+-s "$groupsource"} --completion @*)"
+		if [ -n "$cleangroup" ]; then
+			options=${options//@"$groupsource":/@}
+		fi
+		;;
+	-s|--groupsource)
+		options=$(cluset --groupsources | sed -e 's/ (default)//')
```
thiell, line 61: To avoid the sed:
```
cluset --groupsources --quiet
```

bash_completion.d/clush
```diff
@@ -5,7 +5,7 @@ _clush()
 {
 	# shellcheck disable=SC2034 # set/used by _init_completion
 	local cur prev words cword split
-	local word options="" skip=argv0 groupsource="" cleangroup=""
+	local word options="" compopts="" skip=argv0 groupsource="" cleangroup=""
```
degremont, line 8: I assume we won't squash all commits when this PR will be merged? If yes, better merge this commit in the original bash_completion commit, as this is a fixup

degremont (commented):
Overall, this is a very nice work! I'm just not convinced by the missing completion for `nodeset`

bash_completion.d/clush
thiell: I avoided that because of different authors (Dominique vs. myself), but now I realized I already pushed another fixup before, so I guess it makes sense to merge this too. Will do.

thiell (approved):

PR #581, opened 2025-02-07
NodeUtils: do not pass stdin to group upcalls

martinetd:
Some commands consume stdin when they shouldn't, but we shouldn't have been giving stdin to upcalls in the first place.

Reported-by: Kilian Cavalotti <kilian@stanford.edu>

replaces #580

### DIFF
--- a/lib/ClusterShell/NodeUtils.py
+++ b/lib/ClusterShell/NodeUtils.py
@@ -46,6 +46,13 @@ import time
 from string import Template
 from subprocess import Popen, PIPE
 
+# compat with python 2.7, import directly above for 3.x
+try:
+    from subprocess import DEVNULL
+except ImportError:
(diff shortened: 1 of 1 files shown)


degremont (approved):
Maybe we should add a line in the right place in documentation to clarify that?

Otherwise, LGTM!

martinetd:
Hmm, documentation doesn't seem to say much about how upcalls are executed currently, I think it's fine as it is.

`Group source upcalls` in doc/sphinx/config.rst could perhaps say something (e.g. "upcalls' standard output is used if the command returns 0, standard input is closed and standard error is shared with the caller"), especially the part about requiring success execution now that I noticed it, I don't know.

degremont:
Actually you're right, this is not that important.

PR #585, opened 2025-02-20
completion: try to complete copy or command

### LINKED ISSUE #584: 1.9.3: File path completion in copy mode
It looks like file path completion doesn't work anymore in 1.9.3, when using copy mode:
```
$ clush -c /tmp/file[TAB][TAB]
```
doesn't complete `/tmp/filename` anymore.

Maybe a side effect of the new bash completions [introduced](https://github.com/cea-hpc/clustershell/pull/563) in 1.9.3?

martinetd:
When using clush -c we should try to complete local files.

When not using -c, if a taget has already been set we can also try to complete a command.
We could do like ssh and try to complete it on any node in the nodeset, but a local completion is probably good enough.

And still complete possible options but only do so if a dash was present, this is similar to what e.g. `ls` completion does

Fixes #584

### DIFF
--- a/bash_completion.d/clush
+++ b/bash_completion.d/clush
@@ -1,16 +1,43 @@
 # clush bash completion
 #
 # to install in /usr/share/bash-completion/completions/ or ~/.local/share/bash-completion/completions/
+_clush_command_or_file() {
+	# undo our nospace setting...
+	compopt +o nospace
+
+	# complete either files (copy mode) or commands (if target set)
+	case "$target_set,$mode" in
+	*,copy)
+		# available since bash-completion 2.12
+		if declare -F _comp_compgen_filedir >/dev/null; then
+			_comp_compgen_filedir
+		else
+			_filedir
+		fi
+		;;
+	1,command)
+		# available since bash-completion 2.12
+		if declare -F _comp_command_offset >/dev/null; then
+			_comp_command_offset "$i"
+		else
+			_command_offset "$i"
+		fi
+		;;
+	esac
+}
+
 _clush()
 {
 	# shellcheck disable=SC2034 # set/used by _init_completion
 	local cur prev words cword split
-	local word options="" compopts="" skip=argv0 groupsource="" cleangroup=""
+	local i word options="" compopts="" skip=argv0 groupsource="" cleangroup=""
+	local mode=command target_set=""
 
 	_init_completion -s -n : || return
 
 	# stop parsing if there had been any non-option before (or --)
-	for word in "${words[@]}"; do
+	for i in "${!words[@]}"; do
+		word="${words[i]}"
 		case "$skip" in
 		"") ;;
 		groupsource)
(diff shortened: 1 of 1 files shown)


bash_completion.d/clush
```diff
@@ -1,16 +1,29 @@
 # clush bash completion
 #
 # to install in /usr/share/bash-completion/completions/ or ~/.local/share/bash-completion/completions/
+_clush_command_or_file() {
+	# undo our nospace setting...
+	compopt +o nospace
+
+	# complete either files (copy mode) or commands (if target set)
+	case "$target_set,$mode" in
+	*,copy) _comp_compgen_filedir;;
```
kcgthb, line 10: Isn't `_comp_compgen_filedir` be a "recent" addition to `bash-completion`? Looks like it's been [introduced](https://github.com/scop/bash-completion/commit/6e8a1546bde205fd7064d9e578f35ec67e2c6840) in 2.12, and I think even el9 only ships `bash-completion` 2.11.
Maybe we need a more backward compatible alternative?

bash_completion.d/clush
```diff
@@ -1,16 +1,29 @@
...
+	1,command) _comp_command_offset "$i";;
```
martinetd, line 11: ... And this as well, renamed from _command_offset to _comp_command_offset in 2.12.0 (itself available since 1.0 so forever)

bash_completion.d/clush
martinetd: oh, good catch!
Looks like that commit just renamed `_filedir` which has been around since forever, but they didn't keep an alias for the old one so I guess we'll need to check..l

It's not too hard to check once at source time and keep the variant to use in a variable, probably better than install time check...

martinetd: Done, the check takes 4~5us on my machine so just checking everytime instead of hogging a variable for this. Tested on debian stable (also bash-completon 2.11) and fedora 41 (bash-completion 2.16)

kcgthb: Thanks @martinetd!

kcgthb:
Thanks for the fix!

I know I'm being a nitpicker here, but two more minor details :)

1. nodeset/group completion works for short options only if there's a space after the option. For instance:
   - `clush -w@ log<TAB>` completes to `clush -w@ login` as expected
   - but `clush -w@log<TAB>` doesn't completes to anything (although `-w@login` is a valid syntax

2. command completion only works when it's placed before filename completion. But if `-c` happens first in the command line, then command completion doesn't work anymore. For instance, this doesn't complete the nodeset:
   ```
   $ clush -c /etc/slurm/slurm.conf -w @log<TAB>
   ```

martinetd:
1. is unrelated to this PR, opened https://github.com/cea-hpc/clustershell/pull/586
2. fixed

In either case, we allow options to be way too complex... WTH does `clush -c file1 -w nodes file2` even work?!

Oh, combining the two patches `clush -c<tab>` doesn't properly complete file names from the start of the files; I'm tempted to say I don't care enough so just skipping... (as this involves standard bash-completion helpers, while we can lie about $cur I don't know how to re-add the `-c` prefix at the start of completions... So you won't be able to strip that space down, sorry :P)
Ah, while I'm on weird stuff things like `-vwnode` also doesn't work. I'll also leave that one as an exercise to the reader...

PR #586, opened 2025-03-04
completion: handle short opts without spaces

martinetd:
Apparently some people want to use clush -wfoo...

(Since it's a different issue than #585 this is open as a different PR, I don't like the merge queue squashing unrelated commits :p Note it should be merged after #585 as I've left the other commit in for easier testing, but there's no conflict so we could rebase if required. Opened as draft because of this.)


thiell:
Rebasing and merging for 1.9.4 release as we've been using this for a while now on Sherlock. Thanks!

PR #605, opened 2026-02-25
Updated documentation #604

cfs-pure:
Updated configuration documentation to make it more explicit that clush.conf is in ConfigParser format and in the RST docs clarified that the default run-time global section as `[Main]`.

Updated references to ReadTheDocs to use both https and the current externally facing hostname.

Regenerated manpages from text sources.

Closes #604

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Documentation
 
 Online documentation is available here:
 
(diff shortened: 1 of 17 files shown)


thiell (approved):
Looks good, thanks!

PR #606, opened 2026-04-04
From setup.py to pyproject

### LINKED ISSUE #587: Fedora Packaging: Stop using deprecated python setup.py build/install command in the spec
Source: RH [Bug 2378535](https://bugzilla.redhat.com/show_bug.cgi?id=2378535) - clustershell: Stop using deprecated python setup.py build/install command in the spec

See also: https://fedoraproject.org/wiki/Changes/DeprecateSetuppyMacros

Pcornat:
I made a `pyproject.toml` file to resolve an issue. I tested only the `pip install -e .` command.
I made a small change to introduce `setuptools-scm` to make the build process get package's version from git tags.

Feel free to do whatever is needed to make it compliant with the project and to solve the issue.

Resolves cea-hpc/clustershell#587


pyproject.toml
```diff
@@ -42,6 +42,7 @@ classifiers = [
 Documentation = "https://clustershell.readthedocs.io/"
 Repository = "https://github.com/cea-hpc/clustershell.git"
 Issues = "https://github.com/cea-hpc/clustershell/issues"
+Changelog = "https://github.com/cea-hpc/blob/master/ChangeLog"
```
thiell: Should be https://github.com/cea-hpc/clustershell/blob/master/ChangeLog

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+keywords = ['clustershell', 'clush', 'clubak', 'nodeset']
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Environment :: Console",
+    "Intended Audience :: System Administrators",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: POSIX :: BSD",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 2.7",
```
thiell: this is inconsistent with `requires-python = ">=3.6"` above, I think it's fine if you just remove it.

pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[build-system]
+requires = ["setuptools >= 77.0.3", "setuptools-scm[toml]"]
```
thiell: It is a strict version requirement for setuptools that won't build on EL9 (v53) or EL10 (v69).

thiell (changes requested):
Hello and thanks for your contribution @Pcornat
Could you explain what doesn't work for you today?

For example, re: RPM build, right now we're using `%pyproject_*` macros are not strictly tied to `pyproject.toml` and they work with the legacy setup.py/setup.cfg combo:
https://src.fedoraproject.org/rpms/clustershell/blob/rawhide/f/clustershell.spec
So https://github.com/cea-hpc/clustershell/issues/587 is actually already fixed without using a `pyproject.toml`

Adding your `pyproject.toml` will likely break the build on EL9 and EL10. That said, we could decide to move forward and include it, handling EL builds separately, if there is a good reason for it.

That said, a minimal `pyproject.toml` with just the `[build-system]` declaration (no `[project]` table, keeping metadata in `setup.py`/`setup.cfg`) could be useful for `pip install -e .` workflows and would work across all platforms without the setuptools version constraints? Would that work for you?

pyproject.toml
Pcornat: I lowered the required version to v53

pyproject.toml
Pcornat: I removed it

pyproject.toml
Pcornat: Thank you, I did not see the error !

Pcornat:
Hello @thiell

As the issue is not closed, I thought the problem was still here. So now I think my contribution is useless (?). Would it be possible to close the issue #587 to know that the problem is yesterday's news ?

Plus, removing the `[project]` part in `pyproject.toml` is impossible as it is mandatory for this format. For the Python foundation, the `pyproject.toml` is aimed to replace the `setup.py`, that is why it cannot work without the `[project]` part.

PR #609, opened 2026-05-04
NodeUtils: add envvars option for group source upcalls

thiell:
Adds an optional envvars key to groups.conf.d source sections. It accepts a space-separated list of KEY=VALUE assignments (shell quoting supported) that are injected into the subprocess environment for all upcall commands of that source. Variables already present in the calling process environment take precedence, allowing per-run overrides without editing the config file.
```
  [ansible]
  envvars: ANSIBLE_INVENTORY=/path/to/inventory
  map: ansible-inventory -i $ANSIBLE_INVENTORY --list ...
```

### DIFF
--- a/doc/man/man5/groups.conf.5
+++ b/doc/man/man5/groups.conf.5
@@ -135,6 +135,13 @@ external command is avoided automatically.
 .B cache_time
 Number of seconds each upcall result is kept in cache, in memory only.
 Default is 3600 seconds. This is useful only for daemons using nodegroups.
+.TP
+.B envvars
+Optional space-separated list of \fIKEY=VALUE\fP environment variable
+assignments made available to all upcall commands of this group source.
+Shell quoting is supported for values containing spaces (eg.
+\fIMY_VAR="some value"\fP). Variables already set in the process environment
+take precedence over values defined here.
 .UNINDENT
 .sp
 When the library executes a group source external shell command, the current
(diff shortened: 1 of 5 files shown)


degremont:
If i understood correctly, this is doing the same as, already supported?
```yaml
[ansible]
map: ansible-inventory -i ${ANSIBLE_INVENTORY:-/path/to/inventory} --list ...
```

I'm trying to figure out what is now posible which was not prior.

degremont:
If the purpose is to do an ansible source, you can already do it:
```ini
[ansible]
map: ANSIBLE_INVENTORY=${ANSIBLE_INVENTORY:-foobar} ansible-inventory --list 2>/dev/null | jq -r --arg g "$GROUP" 'def r(d;g): (d[g].hosts//[])[], ((d[g].children//[])[] as $c | r(d;$c)); [r(.;$g)][]'
all: ANSIBLE_INVENTORY=${ANSIBLE_INVENTORY:-foobar} ansible-inventory --list | jq -r '._meta.hostvars | keys[]'
list: ANSIBLE_INVENTORY=${ANSIBLE_INVENTORY:-foobar} ansible-inventory --list 2>/dev/null | jq -r '[keys[] | select(. != "_meta" and . != "all")][]'
```

Actually, I would start with pushing a PR for Ansible like above (except if this is not working for you), and then we can see if this PR is still useful?

thiell:
Ok to switch and submit the ansible PR first with your method. Still, something that could be nice is a mechanism  for users to edit a single variable that can be used by a group source. Today, you have to edit the path for each upcall in the case of Ansible. It would be nice to say: edit your inventory path here, once, and you're good to go.

PR #610, opened 2026-05-05
conf: add Ansible inventory group source example

thiell:
Provides a groups.conf.d example for using Ansible inventory as a ClusterShell group source via ansible-inventory and jq. Supports map (with recursive child group resolution), mapall, all, and list upcalls.

`ansible-inventory --list` always dumps the whole inventory regardless of the queried group, so the per-group `map` upcall already pays the full-dump cost. The `mapall` upcall (#613) exploits this to resolve every group in a single call — e.g. `nodeset -ll` drops from one `ansible-inventory` invocation per group to one. Group names containing `:` or whitespace are skipped from mapall and resolved through `map`.

The inventory path is taken from `$ANSIBLE_INVENTORY` if set, otherwise from the `/path/to/inventory` placeholder, so it can be exported once in the environment (supporting comma-separated inventories) or edited directly in the file.

### DIFF
--- a/conf/groups.conf.d/ansible.conf.example
+++ b/conf/groups.conf.d/ansible.conf.example
@@ -1,0 +1,25 @@
+# Additional ClusterShell group source config file
+#
+# Please see `man 5 groups.conf` for further details.
+#
+
+# Ansible inventory bindings
+#
+# Requirements: ansible-core (provides ansible-inventory), jq
+#
+# Replace /path/to/inventory with your inventory path or directory,
+# or set ANSIBLE_INVENTORY in your environment to override it.
+#
+[ansible]
+
+# Resolve group members, recursing into child groups
+map: ANSIBLE_INVENTORY="${ANSIBLE_INVENTORY:-/path/to/inventory}" ansible-inventory --list 2>/dev/null | jq -r --arg g "$GROUP" 'def r(d;g): (d[g].hosts//[])[], ((d[g].children//[])[] as $c | r(d;$c)); r(.;$g)'
+
+# Resolve all groups in a single call, recursing into child groups
+mapall: ANSIBLE_INVENTORY="${ANSIBLE_INVENTORY:-/path/to/inventory}" ansible-inventory --list 2>/dev/null | jq -r 'def r(d;g): (d[g].hosts//[])[], ((d[g].children//[])[] as $c | r(d;$c)); . as $d | keys[] | select(. != "_meta" and . != "all" and (test("[[:space:]:]") | not)) | "\(.): \([r($d;.)] | join(" "))"'
+
+# All inventory hosts
+all: ANSIBLE_INVENTORY="${ANSIBLE_INVENTORY:-/path/to/inventory}" ansible-inventory --list 2>/dev/null | jq -r '.[].hosts // empty | .[]'
+
+# All groups
+list: ANSIBLE_INVENTORY="${ANSIBLE_INVENTORY:-/path/to/inventory}" ansible-inventory --list 2>/dev/null | jq -r 'keys[] | select(. != "_meta" and . != "all")'
--- a/doc/sphinx/config.rst
+++ b/doc/sphinx/config.rst
@@ -712,6 +712,54 @@ Example of use with :ref:`cluset-tool`::
 
 .. highlight:: text
 
+.. _group-ansible-bindings:
+
+Ansible inventory group bindings
+"""""""""""""""""""""""""""""""""
+
+Enable Ansible inventory group bindings by renaming the example configuration
+file usually installed as
+``/etc/clustershell/groups.conf.d/ansible.conf.example`` to ``ansible.conf``.
+
+**Requirements**: ``ansible-core`` (provides the ``ansible-inventory`` command)
+and ``jq``.
+
+The section **ansible** defines a group source backed by Ansible inventory.
+Each upcall command uses ``ANSIBLE_INVENTORY`` as an inline environment variable
+prefix so that multiple inventory sources (comma-separated paths) are supported.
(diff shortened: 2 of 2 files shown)


conf/groups.conf.d/ansible.conf.example
```diff
@@ -0,0 +1,22 @@
...
+#
+# Requirements: ansible-core (provides ansible-inventory), jq
+#
+# Replace /path/to/inventory with your inventory path or directory,
+# or set ANSIBLE_INVENTORY in your environment to override it.
+#
+[ansible]
+
+# Resolve group members, recursing into child groups
+map: ansible-inventory -i "${ANSIBLE_INVENTORY:-/path/to/inventory}" --list 2>/dev/null | jq -r --arg g "$GROUP" 'def r(d;g): (d[g].hosts//[])[], ((d[g].children//[])[] as $c | r(d;$c)); [r(.;$g)] | unique[]'
```
degremont: In my example i put ANSIBLE_INVENTORY as a variable at the beginning of the line, this is needed due to how `ansible-inventory` handles `-i` . Also, `unique` is not needed as NodeSet will handle that.

Ansible has a weird handling of ANSIBLE_INVENTORY and `-i`. They do not work the same way. 
If you specify `ANSIBLE_INVENTORY=foo,bar` as a env variable, this will look for 2 inventories: `foo` and `bar`. But if you put `-i foo,bar` it will look for host `foo` and `bar`. So `-i $ANSIBLE_INVENTORY` is actually wrong :)

conf/groups.conf.d/ansible.conf.example
```diff
@@ -0,0 +1,22 @@
...
+
+# All inventory hosts
+all: ansible-inventory -i "${ANSIBLE_INVENTORY:-/path/to/inventory}" --list 2>/dev/null | jq -r '._meta.hostvars | keys | sort[]'
```
degremont: same here, env variable at the begining and `sort` is not needed.

conf/groups.conf.d/ansible.conf.example
```diff
@@ -0,0 +1,22 @@
...
+
+# All groups
+list: ansible-inventory -i "${ANSIBLE_INVENTORY:-/path/to/inventory}" --list 2>/dev/null | jq -r '[keys[] | select(. != "_meta" and . != "all")] | sort[]'
```
degremont: ditto

doc/sphinx/config.rst
```diff
@@ -680,6 +680,47 @@ Example of use with :ref:`cluset-tool`::
...
+and ``jq``.
+
+The section **ansible** defines a group source backed by Ansible inventory.
+The inventory path defaults to the value of ``$ANSIBLE_INVENTORY`` if set in
+the environment, or falls back to the path defined in the configuration file::
+
+    [ansible]
+
+    # Resolve group members, recursing into child groups
+    map: ansible-inventory -i "${ANSIBLE_INVENTORY:-/path/to/inventory}" --list 2>/dev/null | jq -r --arg g "$GROUP" 'def r(d;g): (d[g].hosts//[])[], ((d[g].children//[])[] as $c | r(d;$c)); [r(.;$g)] | unique[]'
```
degremont: i'm not sure duplicating the config file in the doc is a good idea, i risk this is not maintained over time. Maybe we just need. If the purpose is to show how the variable is used, may be just putting the `ansible-inventory` part is enough, and simpler to maintain.

degremont (changes requested):

conf/groups.conf.d/ansible.conf.example
thiell: Good to know! I'm new with Ansible so this is very helpful!

doc/sphinx/config.rst
thiell: Agreed thx

thiell:
Reworked per review: inline `ANSIBLE_INVENTORY` prefix on each upcall (no `-i`), dropped `unique`/`sort` (NodeSet dedups), and corrected the doc example output; `map:` stays recursive since a parent group lists only its children, not their hosts.

degremont (approved):
We might change it to include `mapall`, depending which patch lands first.

degremont (commented):
Looks good, but i need to do a real test on system, which i do not have handy right now

thiell:
I tested on a EL8 VM with both .ini and .yml inventory files. At first glance, it looks great, but I noticed something that _could_ be a problem especially in the cloud:

> The map, mapall, and all upcalls emit hosts from the hosts[] arrays of ansible-inventory --list — i.e. each host's **inventory_hostname**. The connection address **ansible_host** (carried in _meta.hostvars[<host>].ansible_host) is never consulted.
> 
> Per the [Ansible inventory docs](https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html), the two are deliberately decoupled:
>   - **ansible_host**: "the resolvable name or IP of the host to connect to, if it is different from the alias you wish to give to it" → may be an FQDN, not only an IP.
>   - **inventory_hostname**: "can also be just an 'alias' or short name for the host" → need not be resolvable. Canonical doc example: jumper ansible_port=5555 ansible_host=192.0.2.50.

That said, `inventory_hostname` is used by `ansible-inventory --graph` so it's probably the best default here, but some users in the cloud could want to use **ansible_host** instead so that `clush` can connect to actual hosts...

I'm thinking at least adding a note about that in the docs, but keeping the current examples (and maybe providing commented examples using **ansible_host**?)

thiell:
Updated patch with a note in the doc about `ansible_host` and also improved mapall's filter (skip group names with ':' or whitespace which are not expressible as 'group: nodes'. Now using POSIX `:space:` which takes care of tab, NL, CR, etc.).

PR #613, opened 2026-05-07
NodeUtils: add mapall upcall to get all group mappings in one call

### LINKED ISSUE #612: Add a new group source upcall to collect the whole config in 1 call
Add a new upcall for group sources (name to be defined, `dump`, `fetch`, ... ?) that will collect all mapping in 1 call.

It is often the case that 1 upcall is costy, could be 1, 2 or more second long, but can dump all mapping in 1 path. (ie: slurm, ansible-inventory, etc...). In our current config, we need to call these upcall 1 for listing the groups, then 1 per groups, which could be very slow).

We will add a new optional upcall.
* If present, `map` is no more mandatory. Minimum requirement is either `map` or this new upcall.
* If available, it will be called first. The expected output will be something like: `group: content`, 1 per line.
* It will fill the cache with this content.
* If `list` is missing, they will be obtained from the cache
* `all` keep the same behavior
* `map` checks the cache first, and then call the upcall, if it exists, on missing group.

* When should we call this one again? (for daemon)
  * When the cache timeout will be reached, entry could be dropped. We can call again if the cache is empty?

thiell:
Group sources backed by Slurm, Ansible inventory, etc. can typically dump every group-to-nodes mapping in a single command, but ClusterShell forced N+1 sub-process invocations (one 'list' followed by one 'map' per group). On large or slow backends this is a meaningful overhead.

This patch introduces an optional 'mapall' upcall whose output is one 'group: nodes' line per group. When defined, ClusterShell runs it once to populate the cache and then serves 'map' and 'list' queries from it. Either 'map' or 'mapall' is required; both may coexist, in which case 'map' acts as a per-group fallback for groups missing from 'mapall'.

API notes:
- `UpcallGroupSource`: `map_upcall` now defaults to `None` (validation is done at config level); `mapall_upcall` is appended at the end of the signature so existing positional callers are unaffected.
- `groups.conf.5` regenerated from `doc/txt/groups.conf.txt`.

Closes #612.

### DIFF
--- a/lib/ClusterShell/NodeUtils.py
+++ b/lib/ClusterShell/NodeUtils.py
@@ -164,11 +164,15 @@ class UpcallGroupSource(GroupSource):
 
     Upcall results are cached for a customizable amount of time. This is
     controlled by `cache_time` attribute. Default is 3600 seconds.
+
+    The optional 'mapall' upcall returns all group-to-nodes mappings in a
+    single call and is then used to serve both 'map' and 'list' queries
+    from the cache.
     """
 
-    def __init__(self, name, map_upcall, all_upcall=None,
+    def __init__(self, name, map_upcall=None, all_upcall=None,
                  list_upcall=None, reverse_upcall=None, cfgdir=None,
-                 cache_time=None):
+                 cache_time=None, mapall_upcall=None):
         GroupSource.__init__(self, name)
         self.verbosity = 0 # deprecated
         self.cfgdir = cfgdir
@@ -176,7 +180,8 @@ class UpcallGroupSource(GroupSource):
 
         # Supported external upcalls
         self.upcalls = {}
-        self.upcalls['map'] = map_upcall
+        if map_upcall:
+            self.upcalls['map'] = map_upcall
         if all_upcall:
             self.upcalls['all'] = all_upcall
         if list_upcall:
@@ -184,6 +189,8 @@ class UpcallGroupSource(GroupSource):
         if reverse_upcall:
             self.upcalls['reverse'] = reverse_upcall
             self.has_reverse = True
+        if mapall_upcall:
+            self.upcalls['mapall'] = mapall_upcall
 
         # Cache upcall data
         if cache_time is None:
@@ -227,9 +234,6 @@ class UpcallGroupSource(GroupSource):
         If `key' is missing, it is added to provided `cache'. Each entry in a
         cache is kept only for a limited time equal to self.cache_time .
         """
-        if not self.upcalls.get(upcall):
-            raise GroupSourceNoUpcall(upcall, self)
-
         # Purge expired data from cache
         if key in cache and cache[key][1] < time.time():
             self.logger.debug("PURGE EXPIRED (%d)'%s'", cache[key][1], key)
@@ -237,6 +241,9 @@ class UpcallGroupSource(GroupSource):
 
         # Fetch the data if unknown of just purged
         if key not in cache:
+            if not self.upcalls.get(upcall):
+                raise GroupSourceNoUpcall(upcall, self)
+
             cache_expiry = time.time() + self.cache_time
             # $CFGDIR and $SOURCE always replaced
             args['CFGDIR'] = self.cfgdir
@@ -245,11 +252,52 @@ class UpcallGroupSource(GroupSource):
 
         return cache[key][0]
 
+    def _populate_from_mapall(self):
+        """
+        Run the optional 'mapall' upcall and fill the map and list caches
+        from its output. No-op if 'mapall' is not defined or its last
+        result is still fresh.
+        """
+        if 'mapall' not in self.upcalls:
+            return
+        cached = self._cache.get('mapall')
+        if cached is not None and cached[1] >= time.time():
+            return
+
+        content = self._upcall_cache('mapall', self._cache, 'mapall')
+        cache_expiry = self._cache['mapall'][1]
+        new_map = {}
+        try:
+            for line in content.splitlines():
+                line = line.strip()
+                if not line:
+                    continue
+                group, sep, nodes = line.partition(':')
+                group = group.strip()
+                if not sep or not group or len(group.split()) > 1:
+                    raise GroupSourceQueryFailed(
+                        "mapall: invalid line %r (expected 'group: nodes')"
+                        % line, self)
+                nodes = nodes.strip()
+                if group in new_map:
+                    # union duplicate group lines, like multi-line map output
+                    nodes = ','.join(n for n in (new_map[group][0], nodes) if n)
+                new_map[group] = (nodes, cache_expiry)
+        except GroupSourceQueryFailed:
+            del self._cache['mapall'] # do not keep unusable output
+            raise
+        self._cache['map'] = new_map
+        self._cache['list'] = (' '.join(new_map), cache_expiry)
+
     def resolv_map(self, group):
         """
         Get nodes from group 'group', using the cached value if
         available.
         """
+        self._populate_from_mapall()
+        if 'map' not in self.upcalls and 'mapall' in self.upcalls:
+            # no map fallback: unknown group resolves to an empty node set
+            return self._cache['map'].get(group, ('',))[0]
         return self._upcall_cache('map', self._cache['map'], group, GROUP=group)
 
     def resolv_list(self):
@@ -257,6 +305,10 @@ class UpcallGroupSource(GroupSource):
         Return a list of all group names for this group source, using
         the cached value if available.
         """
+        self._populate_from_mapall()
+        if 'list' not in self.upcalls and 'mapall' in self.upcalls:
+            # no list fallback: read mapall result directly (cache_time 0)
+            return self._cache['list'][0]
         return self._upcall_cache('list', self._cache, 'list')
 
     def resolv_all(self):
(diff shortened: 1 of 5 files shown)


lib/ClusterShell/NodeUtils.py
```diff
@@ -166,24 +166,31 @@ class UpcallGroupSource(GroupSource):
...
     """
 
-    def __init__(self, name, map_upcall, all_upcall=None,
-                 list_upcall=None, reverse_upcall=None, cfgdir=None,
-                 cache_time=None):
+    def __init__(self, name, map_upcall=None, all_upcall=None,
+                 list_upcall=None, reverse_upcall=None, fetch_upcall=None,
+                 cfgdir=None, cache_time=None):
         GroupSource.__init__(self, name)
+        if not map_upcall and not fetch_upcall:
```
degremont: This check is usually done in GroupResolverConfig.

lib/ClusterShell/NodeUtils.py
```diff
@@ -661,25 +714,30 @@ def _sources_from_cfg(self, cfg, cfgdir):
                 # Support grouped sections: section1,section2,section3
                 for srcname in section.split(','):
                     if srcname != self.SECTION_MAIN:
-                        # only map is a mandatory upcall
-                        map_upcall = cfg.get(section, 'map', raw=True)
-                        all_upcall = list_upcall = reverse_upcall = ctime = None
+                        # at least one of 'map' or 'fetch' is mandatory;
+                        # validation is performed in UpcallGroupSource.__init__
```
degremont: Why validating over there, whereas it is done here for missing map call?

tests/NodeSetGroupTest.py
```diff
@@ -1375,8 +1375,13 @@ def __init__(self, name, data):
         reverse_upcall = None
         if 'reverse' in data:
             reverse_upcall = 'fake_reverse'
-        UpcallGroupSource.__init__(self, name, "fake_map", all_upcall,
-                                   list_upcall, reverse_upcall)
+        fetch_upcall = None
+        if 'fetch' in data:
+            fetch_upcall = 'fake_fetch'
+        map_upcall = "fake_map" if 'map' in data else None
```
degremont: (nitpick) use the same pattern than the other upcalls. That makes it clearer that this is just the same behavior.

lib/ClusterShell/NodeUtils.py
degremont: If we move this check under `if key not in cache:` that allows calling this function just to read from cache (that will be populated by fetch)

lib/ClusterShell/NodeUtils.py
```diff
@@ -245,18 +253,63 @@ def _upcall_cache(self, upcall, cache, key, **args):
...
+                raise GroupSourceQueryFailed(
+                    "fetch: invalid line %r (expected 'group: nodes')" % line,
+                    self)
+            group, nodes = line.split(':', 1)
+            group = group.strip()
+            if not group:
+                raise GroupSourceQueryFailed(
+                    "fetch: empty group name in line %r" % line, self)
+            new_map[group] = (nodes.strip(), cache_expiry)
+        self._cache['map'] = new_map
```
degremont, line 289: Also fill the `list` cache here.

lib/ClusterShell/NodeUtils.py
```diff
@@ -245,18 +253,63 @@ def _upcall_cache(self, upcall, cache, key, **args):
...
+        self._cache['fetch_expiry'] = cache_expiry
+
     def resolv_map(self, group):
         """
         Get nodes from group 'group', using the cached value if
         available.
         """
+        self._populate_from_fetch()
+        if 'map' not in self.upcalls:
```
degremont: with the changes above, we can now just call `_upcall_cache()` to retrieve from the cache (but keep the _populate_from_fetch() above)

lib/ClusterShell/NodeUtils.py
```diff
@@ -245,18 +253,63 @@ def _upcall_cache(self, upcall, cache, key, **args):
...
+                return ''
+            return entry[0]
         return self._upcall_cache('map', self._cache['map'], group, GROUP=group)
 
     def resolv_list(self):
         """
         Return a list of all group names for this group source, using
         the cached value if available.
         """
+        if 'list' not in self.upcalls and 'fetch' in self.upcalls:
```
degremont: same here

lib/ClusterShell/NodeUtils.py
```diff
@@ -245,18 +253,63 @@ def _upcall_cache(self, upcall, cache, key, **args):
...
 
+    def _populate_from_fetch(self):
+        """
+        Run the optional 'fetch' upcall and populate the map cache from
+        its output. No-op if 'fetch' is not configured or the previous
+        fetch is still fresh.
+        """
+        if 'fetch' not in self.upcalls:
+            return
+        if time.time() < self._cache['fetch_expiry']:
```
degremont: I think we can call _upcall_cache() for fetch, that will help factorize code.
Then, check if it has expired.
```python
if 'fetch' not in self.upcalls:
   return

content = self._upcall_cache('list', self._cache, 'list')

if self._cache['fetch'][1] > time.time():
  return

# Parse content, and fill self._cache['map'] and self._cache['list']
# no need for hackish field self._cache['fetch_expiry']

```

degremont (changes requested):
Thanks for the PR. Several comments to keep the code changes in the same philosophy than the current code.

I propose `gather` or `collect` instead of `fetch` (which sounds like map actually).

lib/ClusterShell/NodeUtils.py
thiell: Done — moved to `GroupResolverConfig._sources_from_cfg`

lib/ClusterShell/NodeUtils.py
thiell: Fixed, validation is done in `GroupResolverConfig._sources_from_cfg` now

tests/NodeSetGroupTest.py
thiell: Done

lib/ClusterShell/NodeUtils.py
thiell: Done

lib/ClusterShell/NodeUtils.py
thiell: Done. Note this gives `mapall` precedence over an explicit `list` upcall (the `list` command is simply never run while the cache is fresh). Related test: `test_mapall_overrides_list`

lib/ClusterShell/NodeUtils.py
thiell: Done for the `map` fallback path. For a mapall-only source, an unknown group resolves to an empty result instead of raising: on a cache miss `_upcall_cache()` would raise `GroupSourceNoUpcall('map')`, which  isn't caught on the `@group` parse path and would print a misleading `"No map upcall defined"` error for what is just an unknown group. Empty resolution is also the contract of the other sources (`GroupSource.resolv_map` returns `self.groups.get(group, '')`.

lib/ClusterShell/NodeUtils.py
thiell: Done — `resolv_list()` is now just `_populate_from_mapall()` + `_upcall_cache('list', ...)`.

lib/ClusterShell/NodeUtils.py
thiell: Done — mapall output is now fetched through `_upcall_cache()` and `fetch_expiry` is gone.

One fix to your sketch: the expiry check must come before calling `_upcall_cache()`, not after. `_upcall_cache()` always returns a fresh entry, so checking expiry after the call would always return early and the output would never be parsed.

Also, on a parse error, the cached entry is dropped before re-raising, so the next query retries the upcall instead of serving an empty cache (test `test_mapall_parse_error_retry`).

thiell:
Agreed that `fetch` reads like a synonym of `map`. I'd avoid `gather` since it's already our vocabulary for output gathering (`clush -b`). I went with `mapall`: it keeps the `map` terminology and says exactly what it does — `map`, for all groups, in one call:

```ini
[slurmpart,sp]
mapall: sinfo -h -o "%R: %N"
```

Branch updated (rebased on master) with the rename and your suggested changes — see inline replies for details.

lib/ClusterShell/NodeUtils.py
```diff
@@ -661,8 +704,18 @@ def _sources_from_cfg(self, cfg, cfgdir):
...
                     if srcname != self.SECTION_MAIN:
-                        # only map is a mandatory upcall
-                        map_upcall = cfg.get(section, 'map', raw=True)
+                        # map or mapall is a mandatory upcall
+                        if not cfg.has_option(section, 'map') and \
+                           not cfg.has_option(section, 'mapall'):
+                            raise GroupResolverConfigError(
+                                "No option 'map' or 'mapall' in section: %r"
+                                % section)
+                        map_upcall = mapall_upcall = None
```
degremont, line 723: (nitpick, just if you refresh the patch) empty line between raise and next line. Keeps variable initialization grouped together, then all the `if cfg.has_option(section, ...):`

lib/ClusterShell/NodeUtils.py
```diff
@@ -245,18 +251,55 @@ def _upcall_cache(self, upcall, cache, key, **args):
...
+        from its output. No-op if 'mapall' is not defined or its last
+        result is still fresh.
+        """
+        if 'mapall' not in self.upcalls:
+            return
+        cached = self._cache.get('mapall')
+        if cached is not None and cached[1] >= time.time():
+            return
+        content = self._upcall_cache('mapall', self._cache, 'mapall')
+        cache_expiry = time.time() + self.cache_time
```
degremont: avoid clock skew.
```
cache_expiry = self._cache['mapall'][1]
```

degremont (commented):
Nice move to `mapall`.
Minor comments, only one is really useful.

lib/ClusterShell/NodeUtils.py
thiell: done

lib/ClusterShell/NodeUtils.py
thiell: done

degremont (approved):

PR #614, opened 2026-05-07
Display: fix line buffering on Python 3.6 (#597)

thiell:
reconfigure() is Py3.7+; on Py3.6 the silent except left stdout/stderr block-buffered when piped. Re-wrap the underlying buffer with line_buffering enabled at construction time. Init-only cost; write hot path unchanged.

Closes #597

### DIFF
--- a/lib/ClusterShell/CLI/Display.py
+++ b/lib/ClusterShell/CLI/Display.py
@@ -25,6 +25,7 @@ CLI results display class
 from __future__ import print_function
 
 import difflib
+import io
 import sys
 import os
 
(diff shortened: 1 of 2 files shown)


degremont (approved):
I'm not a specialist, but LGTM. Just, did you manually validate that this is working better on 3.6 with this patch? What was the behavior with this fix?

thiell:
Ok, confirmed that the patch fixes the behavior with python 3.6.

Example on a Rocky Linux 8.10 VM:

```
$ python3 --version
Python 3.6.8
```

With stock version 1.9.3, here installed as rpm via EPEL:
```
sthiell@r8-build:~$ clush --version
clush 1.9.3
sthiell@r8-build:~$ clush -w localhost 'for i in A B C; do echo $i; sleep 1; done' | awk '{ print strftime("[%H:%M:%S]"), $0; fflush() }'
[19:46:21] localhost: A
[19:46:21] localhost: B
[19:46:21] localhost: C
```

With the patch, the messages now arrive every second, like with Python 3.7+:
```
$ clush -w localhost 'for i in A B C; do echo $i; sleep 1; done' | awk '{ print strftime("[%H:%M:%S]"), $0; fflush() }'
[19:46:43] localhost: A
[19:46:44] localhost: B
[19:46:45] localhost: C
```

thiell:
@degremont reduced the length of a comment and added a clush-level test for that, let me know what you think

degremont (approved):

PR #615, opened 2026-05-08
Tree: ev_pickup fires only when the command can no longer be rerouted (#594)

### LINKED ISSUE #594: Improve ev_pickup events in tree mode on reroute
Lower priority but would be nice to fix `ev_pickup` events when a gateway is down and messages are rerouted in tree mode.

See test `test_tree_run_abort_on_start` in `TreeWorkerTest.py` where the check for `ev_pickup` is currently disabled.

The problem is in `Worker/Tree.py`:

```python
    def _check_ini(self):
        self.logger.debug("TreeWorker: _check_ini (%d, %d)", self._start_count,
                          self._child_count)
        if self.eh is not None and self._start_count >= self._child_count:
            # this part is called once
            self.eh.ev_start(self)
            # Blindly generate pickup events: this could maybe be improved, for
            # example, generated only when commands are sent to the gateways
            # or for direct targets, using MetaWorkerEventHandler.
            for node in self.nodes:
                _eh_sigspec_invoke_compat(self.eh.ev_pickup, 2, self, node)
```
In addition the the problems highlighted by the inline comment, the call to `ev_start()` just before might have aborted the worker, and so ideally `ev_pickup` should not be generated in that case.

thiell:
Previously ev_pickup fired from _check_ini for every node right after ev_start, regardless of whether the command had actually been sent (with an unreachable gateway, the CTL may sit in _sendq forever and the target is eventually rerouted to a working gateway) or whether the worker had been aborted. Reroutes also double-fired pickup for
the same target.

Move emission to the actual send sites:
- direct path: MetaWorkerEventHandler.ev_pickup hook (Worker._on_start  fires once per child node-key)
- gateway path: new PropagationChannel._send_ctl helper invoked once  per actual channel.send. By construction _send_ctl runs only after the channel has reached setup=True (either via the immediate-send path which requires it, or via send_dequeue after CFG-ACK), and reroute via _relaunch is gated on setup=False -- so a target whose ev_pickup has fired cannot be rerouted.

write() and set_write_eof() keep the no-pickup path: subsequent traffic on an already-picked-up node must not double-fire.

_emit_pickup gates on _aborted and buffers events before ev_start. The 1:1 ev_pickup-per-node invariant is held structurally by the call sites. Reroute is signaled only by _ev_routing.

Closes #594

### DIFF
--- a/lib/ClusterShell/Worker/Tree.py
+++ b/lib/ClusterShell/Worker/Tree.py
@@ -55,6 +55,13 @@ class MetaWorkerEventHandler(EventHandler):
         self.logger.debug("MetaWorkerEventHandler: ev_start")
         self.metaworker._start_count += 1
 
+    def ev_pickup(self, worker, node):
+        """
+        Called for each node a child worker has just picked up.
+        Propagate the event to the meta worker.
+        """
+        self.metaworker._emit_pickup(node)
+
     def ev_read(self, worker, node, sname, msg):
         """
         Called to indicate that a worker has data to read.
@@ -163,6 +170,10 @@ class TreeWorker(DistantWorker):
         self._target_count = 0
         self._has_timeout = False
         self._started = False
+        self._initialized = False
+        self._aborted = False
+        # buffered pickup events queued during _launch() (before ev_start)
+        self._pending_pickups = []
 
         if self.command is None and self.source is None:
             raise ValueError("missing command or source parameter in "
@@ -365,6 +376,7 @@ class TreeWorker(DistantWorker):
         self.logger.debug('_copy_remote: tar cmd: %s', cmd)
 
         pchan = self.task._pchannel(gateway, self)
+        # ev_pickup is fired by PropagationChannel._send_ctl after the send
         pchan.shell(nodes=targets, command=cmd, worker=self, timeout=timeout,
                     stderr=self.stderr, gw_invoke_cmd=self.invoke_gateway,
                     remote=self.remote)
@@ -381,6 +393,7 @@ class TreeWorker(DistantWorker):
         self.gwtargets.setdefault(str(gateway), set()).update(targets)
 
         pchan = self.task._pchannel(gateway, self)
+        # ev_pickup is fired by PropagationChannel._send_ctl after the send
         pchan.shell(nodes=targets, command=cmd, worker=self, timeout=timeout,
                     stderr=self.stderr, gw_invoke_cmd=self.invoke_gateway,
                     remote=self.remote)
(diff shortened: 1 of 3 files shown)


lib/ClusterShell/Worker/Tree.py
```diff
@@ -492,17 +518,41 @@ def _on_routing_event(self, arg):
...
             self.eh.ev_start(self)
-            # Blindly generate pickup events: this could maybe be improved, for
-            # example, generated only when commands are sent to the gateways
-            # or for direct targets, using MetaWorkerEventHandler.
-            for node in self.nodes:
-                _eh_sigspec_invoke_compat(self.eh.ev_pickup, 2, self, node)
+            # Flush pickup events buffered during _launch(). If ev_start
+            # aborted the worker, no pickups are emitted (#594).
+            pending, self._pending_pickups = self._pending_pickups, []
+            if not self._aborted:
```
degremont: (minor)
looks to be:
```python
for node in pending:
  if self._aborted:
    ...
```
i do not see the need for checking `aborted` twice.

degremont (approved):
Harder to review, trusting you here.

lib/ClusterShell/Worker/Tree.py
thiell: thx, done

degremont:
LGTM, how show case the fix in real life? 
Do you have an example of behavior with/without the fix? Idk, maybe a clush command in tree mode, with a GW down? What would happen, the active commands would be wrong without the fix?

degremont:
From AI review:
> To address it cleanly, move gateway-routed ev_pickup emission from TreeWorker._execute_remote() / _copy_remote() to the point where the propagation channel actually sends the queued control message.

> On top of the PR patch, I’d change PropagationChannel so shell() emits pickups only when the control message is really sent, then remove the early pickup loops from TreeWorker.

Rough patch shape:

```diff
diff --git a/lib/ClusterShell/Propagation.py b/lib/ClusterShell/Propagation.py
@@
-    def send_queued(self, ctl):
+    def _send_ctl(self, ctl, pickup_worker=None, pickup_nodes=None):
+        self.send(ctl)
+        if pickup_worker is not None:
+            for node in pickup_nodes:
+                pickup_worker._emit_pickup(node)
+
+    def send_queued(self, ctl, pickup_worker=None, pickup_nodes=None):
         """helper used to send a message, using msg queue if needed"""
         if self.setup and not self._sendq:
             # send now if channel is setup and sendq empty
-            self.send(ctl)
+            self._send_ctl(ctl, pickup_worker, pickup_nodes)
         else:
             self.logger.debug("send_queued: %d", len(self._sendq))
-            self._sendq.appendleft(ctl)
+            self._sendq.appendleft((ctl, pickup_worker, pickup_nodes))
 
     def send_dequeue(self):
         """helper used to send one queued message (if any)"""
         if self._sendq:
-            ctl = self._sendq.pop()
+            ctl, pickup_worker, pickup_nodes = self._sendq.pop()
             self.logger.debug("dequeuing sendq: %s", ctl)
-            self.send(ctl)
+            self._send_ctl(ctl, pickup_worker, pickup_nodes)
@@
-        self.send_queued(ctl)
+        self.send_queued(ctl, worker, nodes)
```
Then in TreeWorker, drop the new pickup loops added by the PR from both _copy_remote() and _execute_remote():

```diff
-        # Pickup is fired once the command is being sent to the gateway
-        # for each target (#594). On reroute, _launch() calls this method
-        # again so a fresh pickup is emitted per (re-)send.
-        for node in targets:
-            self._emit_pickup(node)

```
Same removal in _execute_remote().

One important adjustment: PropagationChannel.write() and set_write_eof() can keep calling send_queued(ctl) with no pickup args. That avoids changing their behavior.

This keeps the PR’s _emit_pickup() buffering/dedup logic intact, but moves the trigger to the channel layer, where we know whether the command left the local process instead of merely being queued for a gateway that may never come up.

Either clarify that this commit is improving but not fully fixing #594 or add the above logic?

thiell:
@degremont excellent review and AI FTW. I have force-pushed a squashed commit that now does exactly that.

And by the way, I also found another simplification when iterating with my AI:

>   One step beyond the AI sketch: I removed `_picked_up_nodes`. With emission
>   gated by `_send_ctl` (only runs when `setup=True`) and reroute gated by
>   `setup=False`, the dedup is structurally unreachable. The 1:1 ev_pickup-
>   per-node invariant is held by the call sites; new
>   `test_tree_run_gw2f1_no_double_pickup_under_reroute` guards it.

Basically this means that with the existing code structure, `_emit_pickup()` is called at most once per node per worker lifetime. The set `_picked_up_nodes ` is not needed anymore.

My AI warns that:

>   This is structural, not enforced by bytecode. It relies on three invariants holding:
> 
>   1. Worker._on_start(key) is called at most once per (child, key) by the engine.
>   2. send_queued puts each CTL through _send_ctl exactly once (queue OR immediate, never both).
>   3. _relaunch only fires when setup=False (i.e., for CTLs that were never sent).
> 
>   If a future refactor breaks any of those, the guarantee breaks. To catch that:
> 
>   - tests/TreeWorkerTest.py::TreeWorkerGW2F1FTest::test_tree_run_gw2f1_no_double_pickup_under_reroute — upstream regression guard with a list-not-set comparison that catches duplicates.
>   - tests/local_tree_coverage/test_emit_pickup_at_most_once.py — local-only call-count probe that monkey-patches _emit_pickup directly.

I tested the new tests before(failed)/after(OK) so I believe they test what we want.


Finally, you asked:
> how show case the fix in real life? Idk, maybe a clush command in tree mode, with a GW down?

The bug is not visible with clush itself (does not rely on `ev_pickup`, for example its progress counter is driven by `ev_hup`), but the problem is visible with a python program using `ev_pickup`.

For reference, the attached script [cs_demo_594.py](https://github.com/user-attachments/files/28085696/cs_demo_594.py) can be used to reproduce this, here are the results (this PR vs. master):

```
  Results

  On HEAD (commit 18533f0, deeper fix)

  Event sequence:
     0. start    None
     1. pickup   127.0.0.2
     2. hup      127.0.0.2
     3. routing  127.0.0.3
     4. pickup   127.0.0.3     ←  AFTER routing
     5. hup      127.0.0.3
     6. close    None

  rerouted target:    127.0.0.3
  routing event idx:  3
  pickup(127.0.0.3) idx: 4

  PASS: pickup fires AFTER routing (deeper fix in place)
  exit=0

  On master (commit c09f827, no PR #594 work)

  Event sequence:
     0. start    None
     1. pickup   127.0.0.2
     2. pickup   127.0.0.3     ←  fires eagerly, BEFORE the working gateway even sent
     3. hup      127.0.0.2
     4. routing  127.0.0.3
     5. hup      127.0.0.3
     6. close    None

  rerouted target:    127.0.0.3
  routing event idx:  4
  pickup(127.0.0.3) idx: 2

  FAIL: pickup fires BEFORE routing (bug present)
  exit=1
```

Please re-approve when you get a chance. Feels like a bit of a battle of AIs, haha. Thx!

degremont (approved):
I think we're good now! Good job iterating on that.

PR #617, opened 2026-05-22
Tree: pass filter='fully_trusted' to tarfile.extractall() (PEP 706)

thiell:
The bare extractall() call emits a DeprecationWarning on Python 3.12+ and will raise TarError on 3.14. Feature-detect via tarfile.data_filter so older Python versions keep the legacy behavior.

### DIFF
--- a/lib/ClusterShell/Worker/Tree.py
+++ b/lib/ClusterShell/Worker/Tree.py
@@ -41,6 +41,11 @@ from ClusterShell.Worker.Exec import ExecWorker
 from ClusterShell.Propagation import PropagationTreeRouter
 
 
+# PEP 706: 'fully_trusted' preserves legacy extractall() (trusted gateway)
+_TAR_EXTRACT_KWARGS = {'filter': 'fully_trusted'} \
+    if hasattr(tarfile, 'data_filter') else {}
+
+
 class MetaWorkerEventHandler(EventHandler):
     """Handle events for the meta worker TreeWorker"""
 
@@ -467,7 +472,7 @@ class TreeWorker(DistantWorker):
                 try:
                     self.logger.debug("%s extracting %d members in dest %s",
                                       node, len(tmptar.getmembers()), self.dest)
-                    tmptar.extractall(path=self.dest)
+                    tmptar.extractall(path=self.dest, **_TAR_EXTRACT_KWARGS)
                 except IOError as ex:
                     self._on_remote_node_msgline(node, ex, 'stderr', gateway)
                 finally:
--- a/tests/TreeWorkerTest.py
+++ b/tests/TreeWorkerTest.py
@@ -589,6 +589,33 @@ class TreeWorkerTest(TreeWorkerTestBase):
         """test tree rcopy: file, direct target, not in topology"""
         self._tree_rcopy_file(NODE_FOREIGN)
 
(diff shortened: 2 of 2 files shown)


tests/TreeWorkerTest.py
```diff
@@ -589,6 +589,13 @@ def test_tree_rcopy_file_foreign(self):
         """test tree rcopy: file, direct target, not in topology"""
         self._tree_rcopy_file(NODE_FOREIGN)
 
+    def test_tree_rcopy_no_tarfile_warning(self):
+        """test tree rcopy: no tarfile warning leaks (PEP 706)"""
```
degremont: I was thinking: ok we need this test only to check this is fine on old Python versions. So, at some point, we can get rid of it. How would we know?
We need to clarify this is a compat check, like (compat for <Py3.12) or something similar

lib/ClusterShell/Worker/Tree.py
```diff
@@ -41,6 +41,11 @@
 from ClusterShell.Propagation import PropagationTreeRouter
 
 
+# PEP 706: pass filter='tar' when supported (Python 3.7.17+/3.12+)
+_TAR_EXTRACT_KWARGS = {'filter': 'tar'} if hasattr(tarfile, 'data_filter') \
```
degremont: how much did you check that `tar` is the default we need? (i did not verify myself)
because it looks like the default is `None`

thiell: I checked and the default (None or filter='data') would change the behavior by stripping setuid flags, etc. Not really what we want for a tool used by sysadmins I guess. The current behavior is closer to filter='tar', which preserves attributes. As the archiving is kind of trusted (a remote clustershell instance is doing it), I think it's fine. We also add additional protection already with --transform in the command I believe. The legacy behavior is really filter='fully_trusted' but it is indeed risky and I tried to find a reason we would need it but I don't see any. So TL;DR; filter='tar' is a little bit safer than now and should not impact the behavior.

tests/TreeWorkerTest.py
thiell: It's more for Python 3.12-3.13 to make sure they don't emit a warning. I will add a comment to clarify this

thiell: For 3.14+, the current code is broken and other tests would fail.

lib/ClusterShell/Worker/Tree.py
degremont: I'm fine using `tar` to ignore the absolute path in tar, but the group bits will be a problem:
> Clear high mode bits (setuid, setgid, sticky) and group/other write bits ([S_IWGRP](https://docs.python.org/3/library/stat.html#stat.S_IWGRP) | [S_IWOTH](https://docs.python.org/3/library/stat.html#stat.S_IWOTH)).

Write bits for group or others is a problem I think. I assume we do not have a test checking we preserve stuff like 0664 or 0666 in the test, but we probably should add one, and that would mean moving to fully_trusted?

tests/TreeWorkerTest.py
```diff
@@ -589,6 +589,14 @@ def test_tree_rcopy_file_foreign(self):
...
```
degremont: (nitpick) add (<3.14 compat) in the test name

degremont (approved):

PR #621, opened 2026-06-10
tests: avoid dest file conflict in tree copy file tests

thiell:
Fixes an occasional CI failure of `test_tree_copy_file_distant2` (`AssertionError: 2 != 0` — unexpected read events).

The two test targets are IP aliases of the same host, so the concurrent untars share the same dest file, which `make_temp_filename()` pre-creates. When both tars try to replace it at the same time, one can occasionally fail with `tar: Cannot open: File exists`, and its error output is counted as read events by the test.

Fix: remove the pre-created dest file before the copy, which closes the race for two concurrent untars.

### DIFF
--- a/tests/TreeWorkerTest.py
+++ b/tests/TreeWorkerTest.py
@@ -175,6 +175,11 @@ class TreeWorkerTestBase(unittest.TestCase):
         teh = TEventHandler()
         srcf = make_temp_file(b'Lorem Ipsum', 'test_tree_copy_file_src')
         dest = make_temp_filename('test_tree_copy_file_dest')
+        # unlink pre-created dest: test targets are IP aliases of the same
+        # host, so concurrent untars of the same dest path occasionally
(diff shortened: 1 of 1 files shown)


thiell:
Some more info about this problem...

First an example of the (rare) test failure before this patch:

```
FAIL: test_tree_copy_file_distant2 (tests.TreeWorkerTest.TreeWorkerTest)
    self.assertEqual(teh.ev_read_cnt, 0)
AssertionError: 2 != 0
```

This is because of tar extracting a file over an existing one. Removing the pre-created file closes the race **for two writers**: the first tar's exclusive create simply succeeds and it never unlinks, so the second tar (which may still hit `EEXIST` on the first one's file) always finds the name, it unlinks it, and continues without error (normal tar behavior). Be careful, with 3+ concurrent writers on one path, two could still race, so I added a comment about that.

PR #623, opened 2026-06-12
conf: add mapall upcalls to Slurm group bindings

thiell:
Adopt the new `mapall` upcall in the Slurm example group bindings, so whole-source operations (`cluset -ll -s slurmuser`, `@su:*`, regroup) run one slurmctld query per cache window instead of one `list` plus one `map` per group.

- `slurmpart`, `slurmjob`, `slurmuser`, `slurmaccount`, `slurmqos`: single command; duplicate group lines (one per job) are merged by the mapall parser.
- `slurmstate`: awk strips node state flag suffixes from the state field only (`tr` on the whole line would also strip `-` from node names).
- `slurmresv`: reservation names containing `:` cannot be represented in `mapall` output and are skipped; they remain resolvable through `map`.
- `doc/sphinx/config.rst` updated to match.

### DIFF
--- a/conf/groups.conf.d/slurm.conf.example
+++ b/conf/groups.conf.d/slurm.conf.example
@@ -8,6 +8,7 @@
 #
 [slurmpart,sp]
 map: sinfo -h -o "%N" -p $GROUP
+mapall: sinfo -h -o "%R:%N"
 all: sinfo -h -o "%N"
 list: sinfo -h -o "%R"
 reverse: sinfo -h -N -o "%R" -n $NODE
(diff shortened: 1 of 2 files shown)


conf/groups.conf.d/slurm.conf.example
```diff
@@ -26,6 +28,7 @@ cache_time: 60
 #
 [slurmstate,st]
 map: sinfo -h -o "%N" -t $GROUP
+mapall: sinfo -h -o "%T:%N" | awk -F: 'NF>1 {gsub(/[*~#!%$@+^-]/,"",$1); print $1":"$2}'
```
degremont: Might be over-engineered. Using the same pattern as the other lines seems good enough

    sinfo -h -o "%T:%N" | tr -d '*~#!%$@+^-'

degremont (approved):
LGTM, just a minor comment.

conf/groups.conf.d/slurm.conf.example
thiell: tr doesn't work in that case, for example:
```
$ echo 'mixed*:gpu-a100-[01-04]' | tr -d '*~#!%$@+^-'
mixed:gpua100[0104]
```

however, sed is better I think:
```
$ echo 'mixed*:gpu-a100-[01-04]' | sed 's/[*~#!%$@+^-]*:/:/'
mixed:gpu-a100-[01-04]
```

degremont (approved):

PR #624, opened 2026-06-12
conf: add mapall upcall to xCAT group bindings

thiell:
Adopt the new `mapall` upcall in the xCAT example group bindings: one `lsdef -t group -i members` call replaces one `lsdef` call per group, for both static and dynamic groups. Group names containing `:` are skipped and remain resolvable through `map`.

`doc/sphinx/config.rst` updated to match.

### DIFF
--- a/conf/groups.conf.d/xcat.conf.example
+++ b/conf/groups.conf.d/xcat.conf.example
@@ -10,6 +10,9 @@
 # list the nodes in the specified node group
 map: lsdef -s -t node "$GROUP" | cut -d' ' -f1
 
+# get all group-to-nodes mappings in one call
+mapall: lsdef -t group -i members | awk '/^Object name: /{g=$3; next} sub(/^ *members=/,""){if (g != "" && g !~ /:/) print g": "$0}'
+
 # list all the nodes defined in the xCAT tables
 all: lsdef -s -t node | cut -d' ' -f1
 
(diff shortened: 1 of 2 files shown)


kcgthb (commented):
Actually, I'm not really sure about this one. 🤔 

`mapall` works great when the cost of resolving all groups is the same, or close, to the cost of resoling one group.

But that's not the case for xCAT. There could be a number of xCAT groups that have little reason to ever be invoked via `clush`, and resolving all members of groups (especially dynamic groups) could be pretty costly on large systems with many groups. And in that case, resolving *all* groups knowing that the vast majority will never be used, doesn't seem like a very significant optimization.

There's a cluster I know where resolving all members of every single group takes much more time than `cache_time` :D 
```
# time lsdef -t group -i members >/dev/null

real    3m32.317s
user    0m0.327s
sys     0m0.079s
```
This is largely due to the fact that resolving members of a dynamic group is an expensive SQL query in the xCAT DB (the membership is... dynamic, not defined statically) and those SQL queries are pretty much serialized, I believe.

So on a system with many dynamic groups (like this:
```
# lsdef -c -t group -i grouptype | awk -F: '{print $NF}' | sort | uniq -c
     31  grouptype=
    272  grouptype=dynamic
```
I'm not sure that a `mapall` call could be a real improvement, pretty much the opposite. :\

degremont:
That's exactly the kind of data we are looking for. `mapall` is designed for use case where the request to get all groups is similar to get few groups (1-3). If the time is especially more expensive, no need to add it.

Do you have other scenarios, where the config is simpler? What is the perf difference in that situation?

Let's drop it. It could always be added later if that useful.

thiell:
Thanks for the feedback! Agreed, that doesn't make sense to use `mapall` for xCAT in that case.

PR #626, opened 2026-06-13
tests: skip pdsh tests when pdsh is not installed

thiell:
Running the test suite on a host without pdsh is currently fatal: in `test_033_worker_pdsh_tty`, clush's tty code path runs the command in a separate task thread; the failed pdsh spawn raises `WorkerError("Cannot run pdsh (error 255)")` there, which reaches `clush_excepthook` and ends in `os._exit(1)` — silently killing the whole unittest process mid-test. Watching the suite log, this looks like a hang (the log freezes at `test_033 ...` with no result; the error message is lost in the captured stderr).

- add a small `which()` helper to tests/TLib.py
- skip with `@unittest.skipIf(which('pdsh') is None, ...)`: the three `TaskDistantPdshTest` classes, `CLIClushTest` `test_032`/`test_033`, and the two `TaskRLimitsTest` WorkerPdsh tests
- CI installs pdsh explicitly (tests.yml "Install pdsh to test WorkerPdsh"), so CI coverage is unchanged; this only fixes runs on dev machines

Verified without pdsh installed: all guarded tests now skip and the suite completes normally; with pdsh present the guards are no-ops.

### DIFF
--- a/tests/TLib.py
+++ b/tests/TLib.py
@@ -18,11 +18,20 @@ except ImportError:
 from io import BytesIO, StringIO
 
 
-__all__ = ['HOSTNAME', 'load_cfg', 'make_temp_filename', 'make_temp_file',
-           'make_temp_dir', 'CLI_main']
+__all__ = ['HOSTNAME', 'which', 'load_cfg', 'make_temp_filename',
+           'make_temp_file', 'make_temp_dir', 'CLI_main']
 
 # Get machine short hostname
 HOSTNAME = socket.gethostname().split('.', 1)[0]
+
+# Replace with shutil.which() once we only support Py3.3+
+def which(program):
+    """Return path of first executable program found in PATH, or None."""
+    for path in os.environ.get('PATH', '').split(os.pathsep):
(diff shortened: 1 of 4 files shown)


tests/TLib.py
```diff
@@ -18,12 +18,20 @@
...
 
-__all__ = ['HOSTNAME', 'load_cfg', 'make_temp_filename', 'make_temp_file',
-           'make_temp_dir', 'CLI_main']
+__all__ = ['HOSTNAME', 'which', 'load_cfg', 'make_temp_filename',
+           'make_temp_file', 'make_temp_dir', 'CLI_main']
 
 # Get machine short hostname
 HOSTNAME = socket.gethostname().split('.', 1)[0]
 
+def which(program):
```
degremont, line 28: Add a comment like:
` # Should be replaced by shutil.which() when we drop everything < Py3.3`

degremont (approved):
minor comment, otherwise LGTM

tests/TLib.py
thiell: done thx!

Issue #356, opened 2017-09-07
nD nodeset folding: add default folding axis


thiell:
By default, clustershell folds node sets in all dimensions. This can actually be quite annoying sometimes as other tools don't support that. For example, Slurm doesn't and requires that the folding axis is the last one:

```
# scontrol reboot_nodes sh-[2,11,13,30]-4
scontrol_reboot_nodes error: Invalid node name specified
```

What about adding an option to defauts.conf (and maybe clush.conf?) in order to specify a default folding axis (already implemented as a CLI option for nodeset/cluset in a5162d17b870dec44596892c5bf326c38cb6a00d)?

I also opened a Slurm ticket at https://bugs.schedmd.com/show_bug.cgi?id=4138

thiell:
Also, the `--axis` CLI option would be nice to have with `clush`.

thiell:
SchedMD says that the last single digit should be enclosed in brackets..... `NodeSet` won't use brackets unless there is more than a single index for the dimension.

```
So you need to enclose the last number in a bracket, for example,

marshall@byu:~/slurm/16.05/byu$ scontrol show hostnames byu[3-4]-[2]
byu3-2
byu4-2
```

degremont:
I see no real justification for Slurm limitation. Except we are lazy doing it :)

Nevertheless, this default folding axis config option makes sense to me. It could be useful for other use cases.

thiell:
Looks like the Slurm folks are considering extending the syntax! Progress!

On our side, I will set the 'default folding axis' as enhancement for 1.9.

thiell:
Setting a default folding axis is already possible in defaults.conf
See: https://clustershell.readthedocs.io/en/latest/config.html?highlight=fold_axis#use-case-slurm
I am renaming this issue to implement `clush --axis` like with `cluset` but with low priority.

thiell:
Fixed by #629

PR #629, opened 2026-06-17
Clush, Clubak: add --axis option to choose nD fold axis

thiell:
Adds `--axis` to `clush` and `clubak`, the per-invocation equivalent of the
`fold_axis` library default, to fold displayed (gathered-header) node sets
along selected nD axis only — useful for tools (e.g. Slurm) that don't parse
all nD node set forms. `nodeset`/`cluset` already had `--axis`.

- `--axis` translated 1-indexed → 0-indexed `DEFAULTS.fold_axis` (clush/clubak)
- translation factored into a shared `CLI.Utils.parse_fold_axis()` helper, also
  adopted by the existing `nodeset`/`cluset` `--axis` (de-duplicated)
- man + sphinx docs (`clush.rst`, `clubak.rst`, Slurm use case in `config.rst`)
- SSH-free tests for both tools (exec worker / stdin)

Implements #356.

### DIFF
--- a/doc/man/man1/clubak.1
+++ b/doc/man/man1/clubak.1
@@ -68,6 +68,9 @@ disable header block and order output by nodes
 .B  \-r\fP,\fB  \-\-regroup
 fold nodeset using node groups
 .TP
+.BI \-\-axis\fB= RANGESET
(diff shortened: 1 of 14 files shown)


degremont (approved):
code wise ok, but are we sure people will need clush --axis ?

thiell:
It's a good point, you're right it's probably rare, but I kind of like that all clu tools have the same ability to override the config default `fold_axis`).

A potential use case is passing the nodesets from `clush -bL` directly to another tool that does not support nD folding, imagine something like that:

```sh
bad=$(clush -bL --axis=-1 -w @gpu \
      "nvidia-smi -q -d ECC | grep 'DRAM Uncorrectable' | grep -qvE ':[[:space:]]+(0|N/A)\$' && echo BAD || echo OK" \
      | awk -F': ' '$2=="BAD"{print $1}')
        
[ -n "$bad" ] && scontrol update nodename="$bad" state=drain reason="GPU DRAM uncorrectable ECC"
```

PR #632, opened 2026-06-23
NodeSet: add index() method

### LINKED ISSUE #631: Add NodeSet index() method
Python `list` has an `index()` method that will return the zero-based index of the first occurrence of the search term in the list. ClusterShell is missing a similar method for NodeSet, but I assume it could be much more efficient than converting to a list.

```
>>> a = NodeSet('test[01-10]')
>>> a[3]
'test04'
>>> list(a).index('test04')
3
>>> a.index('test04')
  File "<stdin>", line 1, in <module>
AttributeError: 'NodeSet' object has no attribute 'index'
```

Does this sound reasonable?

thiell:
Add `NodeSet.index(node)`, the inverse of `NodeSet[i]`: it returns the
zero-based position of a node, or raises `ValueError` if the node is not
in the nodeset (like `list.index()`).

    >>> NodeSet("node[10-49]").index("node12")
    2

Like `NodeSet[i]`, it walks the nodeset, so it is O(n), not O(1) — but it
avoids building a temporary list. A faster path for sets with many
patterns is possible, but would mean inverting the nD range folding, so
it is left out for now.

Closes #631

### DIFF
--- a/lib/ClusterShell/NodeSet.py
+++ b/lib/ClusterShell/NodeSet.py
@@ -479,6 +479,23 @@ class NodeSetBase(object):
         else:
             raise TypeError("NodeSet indices must be integers")
 
+    def index(self, other):
+        """
+        Return the zero-based index of a single node in the nodeset, the
+        inverse of __getitem__().
+
+        :raises ValueError: node is not contained in the nodeset
+        """
+        self._binary_sanity_check(other)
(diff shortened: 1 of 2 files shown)


degremont:
I could not update this PR, so I pushed a new one: #633

thiell:
Obsoleted by #633

PR #633, opened 2026-06-23
NodeSet, RangeSet: add list-like index() method (#631)

### LINKED ISSUE #631: Add NodeSet index() method

degremont:
Implements NodeSet.index() (and RangeSet/RangeSetND.index()). Alternative to #632.

Add an optimized index() method to NodeSet, RangeSet and RangeSetND that returns the position of an element following iteration order and raises ValueError when it is absent, like the index() method of the list type (including the optional start/stop search window for NodeSet).

The NodeSet implementation does not expand the whole set: it only sums the size of the patterns preceding the searched node and locates it within its own pattern, delegating the per-pattern lookup to RangeSet/RangeSetND.index().

This version, compares to #632 is more optimized, where this would make a performance difference for huge NodeSet length.

Closes #631.

### DIFF
--- a/lib/ClusterShell/RangeSet.py
+++ b/lib/ClusterShell/RangeSet.py
@@ -30,6 +30,24 @@ from functools import reduce
 from itertools import product
 from operator import mul
 
+# Python 3 compatibility
+try:
+    basestring
+except NameError:
+    basestring = str
+
+
+def _normalized_index_bounds(length, start, stop):
+    """Return list.index()-like normalized start/stop bounds."""
+    if start < 0:
+        start = max(0, length + start)
+    if stop is None:
+        stop = length
+    elif stop < 0:
+        stop = max(0, length + stop)
+    return start, stop
+
+
 __all__ = ['RangeSetException',
            'RangeSetParseError',
            'RangeSetPaddingError',
@@ -281,12 +299,18 @@ class RangeSet(set):
         the object is empty, in that case it will return 0."""
         return int(len(self) > 0)
 
+    @staticmethod
+    def _sortkey(elem):
+        """Sort key used to order RangeSet elements (mixed padding support):
+        sort by both string length and index."""
+        if elem.startswith('-'):
+            return (-len(elem), int(elem))
+        return (len(elem), elem)
+
     def _sorted(self):
         """Get sorted list from inner set."""
         # For mixed padding support, sort by both string length and index
-        return sorted(set.__iter__(self),
-                      key=lambda x: (-len(x), int(x)) if x.startswith('-') \
-                                    else (len(x), x))
+        return sorted(set.__iter__(self), key=self._sortkey)
 
     def __iter__(self):
         """Iterate over each element in RangeSet, currently as integers, with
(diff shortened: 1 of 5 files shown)


lib/ClusterShell/RangeSet.py
```diff
@@ -509,6 +521,30 @@ def __getitem__(self, index):
...
+        Return the zero-based index of element in the sorted RangeSet.
+
+        This is the reverse operation of :meth:`RangeSet.__getitem__` and
+        behaves like the ``index()`` method of the ``list`` type. The element
+        may be given as a string (zero-padding is then significant) or as an
+        integer.
+
+        :raises ValueError: the element is not present in the RangeSet
+        """
+        if isinstance(elem, str):
```
thiell: (py2.7 compat) change to `if isinstance(elem, basestring):` to fix this potential use case in py2.7:

```
  >>> u"08" in RangeSet("08-09")        # True
  >>> RangeSet("08-09").index(u"08")    # py2.7
  ValueError: 08 is not in RangeSet
```

That makes `index()` as tolerant as `__contains__` and consistent with `RangeSetND.index()`; on py3 it's a no-op (`basestring` is `str`).

thiell (changes requested):
Thanks for proposing this! This seems to work well!!

Two more things:
1. `start`/`stop` are on `NodeSet.index()` but not `RangeSet/RangeSetND.index()`, I assume it's OK as optional, just checking that you are well aware of this:
```
  # NodeSetBase / NodeSet  — have the window
  def index(self, other, start=0, stop=None): ...

  # RangeSet               — no window
  def index(self, elem): ...

  # RangeSetND             — no window
  def index(self, elem): ...
```

2. Add "Closes #631" to the PR body. It looks like GH uses the PR body for auto-closing now.

lib/ClusterShell/RangeSet.py
```diff
@@ -509,6 +533,37 @@ def __getitem__(self, index):
...
+        integer. The optional `start` and `stop` arguments restrict the search
+        to the matching subsequence, and may be negative (counted from the
+        end).
+
+        :raises ValueError: the element is not present in the RangeSet
+        """
+        if isinstance(elem, basestring):
+            key = elem
+        else:
+            key = "%d" % int(elem)
```
thiell: That line could lead to unexpected behavior:

```
RangeSet("0-9").index(3.5)   # -> 3  (but 3.5 not in set)
```

* `list.index(3.5)` raises
* `RangeSetND.index` uses `"%s" % e` and rejects these, so there is a mismatch between 1D and nD here

Recommend using `key = str(elem)` instead to match `RangeSet.__contains__`

thiell (changes requested):
Just one small thing, otherwise that looks very good. Want to squash into one commit? (We can't squash easily after the fact with the merge queue we're using)

thiell (approved):

PR #637, opened 2026-06-25
Nodeset: add --index command to cluset/nodeset

thiell:
Add a `--index NODE` command to `cluset`/`nodeset` that outputs the zero-based position of a node in the resulting node set. It is the command-line front-end to the `NodeSet.index()` / `RangeSet.index()` methods (#631) and the inverse of `-I/--slice`. Suggested by @mattaezell in #631.

```
$ cluset --index node5 node[0-9]
5
$ cluset -R --index 18 1,5,18-31
2
```

A node that is not in the set prints an error on stderr and exits non-zero (like `list.index()`), so it doubles as a membership test:

```
$ cluset --index $(hostname) $SLURM_JOB_NODELIST   # rank on stdout, or exit 1
```

Notes:
- It is a command (mutually exclusive with `-c`/`-e`/`-f`/...), computed on the final set, so it honors `-x`/`-i`/`-X`, `-I/--slice` and stdin, and works in RangeSet mode (`-R`).
- `--index` combined with `--pick` is rejected, since random sampling would make the position meaningless.
- The position follows ClusterShell's own ordering (same as `-e/--expand`); the docs cover the non-obvious cases (alphabetical pattern order, multidimensional flattening) and warn it is not guaranteed to match a resource manager's rank such as Slurm's `$SLURM_NODEID`.

See #631.

### DIFF
--- a/bash_completion.d/cluset
+++ b/bash_completion.d/cluset
@@ -40,7 +40,7 @@ _cluset()
 	done
 
 	case "$prev" in
-	-c|--count|-e|--expand|-f|--fold|\
+	-c|--count|-e|--expand|-f|--fold|--index|\
 	-x|--exclude|-i|--intersection|-X|--xor)
 		case "$cur" in
 		*:*)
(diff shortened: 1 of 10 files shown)


degremont:
Looks good. Could you also add to the docs that the order is not guaranteed to be stable across versions, especially for multi dimensional NodeSet.

Issue #226, opened 2014-02-02
Add deprecation warning where needed


degremont:
A couple of method and classes are deprecated in ClusterShell.
We need to add call to `warnings.warn` where needed.

The next release will be able to drop this code.

degremont:
Partially done by #271

thiell:
What remaining methods/classes have to be deprecated in 1.8?

degremont:
Maybe current code is enough, I'm listing some deprecated features here:
* WorkerBadArgumentError (1.4)
* WorkerSimple (1.7, still used by WorkerOpen (as of 1.8))
* GroupResolver.set_verbosity() (1.7)
* Worker.last_*() (1.6)
* NodeSet.STD_GROUP_RESOLVER (1.5)
* NodeSet.NOGROUP_RESOLVER (1.5)


* topology.conf: "Main" section (1.7)

thiell:
Thx. Sounds good to me.

I say we remove the first two in 2.0.

There is no rush for the TODO items. Let's do that later.

Let's work on getting rid of `WorkerSimple` in 1.9 as much as possible, so we can remove it in 2.0.

thiell:
- removed all `Worker.last_*()` methods in #495
- added deprecation warnings when old EventHandler method signatures are still used in #358 
- deprecating Worker.current_* in the documentation also in #358

PR #639, opened 2026-06-27
Add deprecation warnings for set_verbosity() and topology [Main] section

thiell:
Add runtime deprecation warnings for two long-deprecated surfaces:

- `GroupResolver.set_verbosity()` — documented deprecated for years but never
  warned at runtime. The `GroupSource.verbosity` value it set is no longer read,
  so the clush/clubak `-d` calls to it were dead no-ops and are removed.
- the pre-v1.7 topology.conf `[Main]` section — silent compat fallback now warns
  (use `[routes]`); stray topology fixtures in the test suite moved
  `[Main]` → `[routes]`.

`DeprecationWarning` is silent for normal CLI users (default filters), visible
only under `-W` or test runners, and never fails the build. The remaining #226
items (constant/class aliases) need py3-only `__getattr__` and are deferred to 1.11.

See #226

### DIFF
--- a/lib/ClusterShell/CLI/Clubak.py
+++ b/lib/ClusterShell/CLI/Clubak.py
@@ -31,7 +31,7 @@ import sys
 
 from ClusterShell.Defaults import DEFAULTS
 from ClusterShell.MsgTree import MsgTree, MODE_DEFER, MODE_TRACE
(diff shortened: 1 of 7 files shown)


degremont (approved):
Funny to see we are adding warnings for stuff we were still using until that patch :)

PR #640, opened 2026-06-27
Task: only terminate Task once (#110)

### LINKED ISSUE #110: testThreadTaskWaitWhenNotStarted failure

thiell:
Aborting a running task with `kill=True` (e.g. via `task_cleanup()`) made
`_terminate()` run twice — from `_resume()`'s `EngineAbortException` handler
(which sets `self._engine = None`) and again from `_thread_start()` — raising
`AttributeError: 'NoneType' object has no attribute 'clear'` on the task thread.

- Fix: return early from `_terminate()` when the engine is already cleared.
- Add a regression test forcing the running-abort race.

The regression test exposed a pre-existing harness wart, fixed in a separate
commit: a `-d`/`--debug` CLI run calls `logging.basicConfig()`, binding a root
handler to `CLI_main`'s captured stderr that is then closed. A later test's
daemon-thread debug log then writes to the closed stream, leaking a traceback
into its captured output — making `testThreadTaskTerminateOnceOnRunningAbort`
flaky in some module subsets. `CLI_main` now snapshots and restores the root
logger's handlers and level around each run.

Closes #110

### DIFF
--- a/lib/ClusterShell/Task.py
+++ b/lib/ClusterShell/Task.py
@@ -962,6 +962,11 @@ class Task(object):
         Abort completion subroutine.
         """
         assert self._quit == True
+
+        # already fully terminated by a prior abort pass (#110)
+        if self._engine is None:
+            return
+
         self._terminated = True
 
         if kill:
--- a/tests/TLib.py
+++ b/tests/TLib.py
@@ -2,6 +2,7 @@
 Unit test library
 """
 
+import logging
 import os
 import socket
 import sys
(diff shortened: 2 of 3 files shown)


degremont (approved):
A fix already available in #480, duplicate?

thiell:
Good catch, #480 is a duplicate indeed, but with a wrong implementation (that's why I never merged it). This PR will supersede it.

degremont (approved):

PR #645, opened 2026-07-04
setup: install ansible.conf.example group source example

thiell:
`ansible.conf.example` ships in the sdist and the 1.10 release notes advertise the Ansible group source upcall, but it was never added to the `groups.conf.d` entry in `setup.py`'s `data_files`, so it installs into no binary package on any distribution. This adds it alongside the existing examples. Distro specs that glob `*.conf.example` pick it up automatically; no packaging changes needed. Intended for 1.10.1.


degremont:
We need to fix the docs and release doc that specify Ansible config example support on 1.10?

degremont (approved):

thiell:
Docs available at https://clustershell.readthedocs.io/en/latest/config.html#group-ansible-bindings

PR #647, opened 2026-07-05
logging: add NullHandler to top-level library logger (#641)

### LINKED ISSUE #641: logging: add a NullHandler to the ClusterShell library logger
## Summary

ClusterShell is used two ways: as command-line tools (`clush`, `nodeset`/`cluset`, `clubak`) and as a Python library that other software embeds. Today the command-line tools configure logging in a way that reaches outside ClusterShell and changes a setting for the *entire* Python program, not just ClusterShell's own code. That has three practical consequences:

- **Embeddability.** When another application imports ClusterShell as a library, ClusterShell can interfere with that application's own logging — a surprising global side effect for downstream tools.
- **Predictability.** Logging behavior differs between our own tools (`clush`, `nodeset`, and `clubak` each do something different), and a couple of small workarounds exist only to mask the side effects.
- **Test reliability.** This global behavior recently caused a flaky test: a logging handler outlived the buffer it was attached to and leaked an error into an unrelated test's output. It was patched in the test harness, but the underlying cause is in the tools, not the tests.

**What we propose:** route all of ClusterShell's logging through its own named logging channel (a dedicated `ClusterShell` "namespace") instead of the shared program-wide one. The library stays silent by default (so importing it has no side effects), and the command-line tools configure only ClusterShell's own channel, through a single shared helper so all tools behave consistently. The useful `clush -d` debug output is preserved exactly as-is, and normal runs stay quiet.

This is a behavior-adjacent cleanup, so it targets a release **after 1.10** rather than the conservative 1.10 line. Relates to #640.

<details>
<summary><strong>Technical details</strong></summary>

### Current behavior

The CLIs configure Python's **root** logger via `logging.basicConfig()`, which is process-global shared state (one handler and level for the whole interpreter):

- `clush` (`lib/ClusterShell/CLI/Clush.py`, around lines 885/888): `basicConfig(level=DEBUG)` when `config.verbosity >= VERB_DEBUG` (i.e. `-d`), else `basicConfig(level=CRITICAL)`.
- `nodeset`/`cluset` (`lib/ClusterShell/CLI/Nodeset.py`, around line 173): `basicConfig(level=DEBUG)` on `options.debug` (`-d`); no `else` branch.
- `clubak` configures no logging at all.

The library does **not** add a `logging.NullHandler` to its top-level `"ClusterShell"` logger (nothing in `lib/ClusterShell/__init__.py` does this). Without one, when ClusterShell is used as a library with no logging configured, its `WARNING`+ records fall through to `logging.lastResort` (Python 3.2+) or trigger `"No handlers could be found for logger ..."` (Python 2.7). The `clush` `else: basicConfig(CRITICAL)` branch effectively exists to paper over this.

The library has ~120 debug emission sites (`Worker/Tree.py`, `Communication.py`, `Gateway.py`, etc.). The `clush -d` debug output is a genuine, useful feature that **must** be preserved.

### Motivating symptom

The #110 regression test `testThreadTaskTerminateOnceOnRunningAbort` (PR #640) was flaky: a `basicConfig` handler got bound to a unit-test capture buffer that was then closed, leaking a `"Logging error"` traceback into a later test's captured output. PR #640 worked around it in `tests/TLib.py` by snapshotting/restoring root logging around each `CLI_main` run — a test-harness workaround, not a fix of the CLI behavior.

### Proposed solution

Adopt the standard Python library/application logging split:

- **Library:** add a `NullHandler` to the top-level `"ClusterShell"` logger in `lib/ClusterShell/__init__.py`.
- **CLIs:** stop calling `logging.basicConfig()`; configure logging **scoped to the `"ClusterShell"` logger namespace** (not root), via one shared helper in `CLI/Utils.py` used by `clush`, `nodeset` (and future tools).
- **Preserve UX:** `-d` / `verbosity >= VERB_DEBUG` → attach a `StreamHandler(sys.stderr)` at `DEBUG` to the `"ClusterShell"` logger; otherwise stay quiet (the `NullHandler` prevents `lastResort`/`"No handlers"` noise); drop the root-`CRITICAL` hack.

`CLI/Utils.py` already hosts shared CLI helpers (`parse_fold_axis`, `human_bi_bytes_unit`, `nodeset_cmpkey`, `bufnodeset_cmpkey`) and is the natural home for a shared `setup_cli_logging()` helper (it does not currently import `logging`). Verbosity plumbing already exists: `CLI/Display.py` defines `VERB_QUIET`/`VERB_STD`/`VERB_VERB`/`VERB_DEBUG`; `clush` uses `config.verbosity >= VERB_DEBUG`, `nodeset` uses `options.debug`.

### Options

- **Option A (recommended):** shared `CLI/Utils.setup_cli_logging(verbosity)` configuring `logging.getLogger("ClusterShell")`; both `clush` and `nodeset` call it. Single source of truth; easy to extend to new tools.
- **Option B:** per-tool config, but target the `"ClusterShell"` logger instead of root. Smaller diff; keeps the current divergence between tools.
- **Option C (minimal stopgap):** keep `basicConfig` but only in the debug branch and drop the root-`CRITICAL` else (rely on the new `NullHandler` for quietness). Still mutates root; least preferred.

### Compatibility

- `clush -d` debug output unchanged; normal `clush`/`nodeset` stay quiet.
- `NullHandler` is available on Python 2.7+, so the library change is py2.7-safe (ClusterShell keeps best-effort py2.7 on the 1.10 line).

### Acceptance criteria

- [ ] No `logging.basicConfig()` or root-logger mutation remains under `CLI/`.
- [ ] Library import adds a `NullHandler` to the `"ClusterShell"` logger and produces no logging output / no `"No handlers"` warning by default.
- [ ] `clush -d` (and `verbosity >= VERB_DEBUG`) still emits library debug output to stderr.
- [ ] Normal `clush`/`nodeset` invocations stay quiet.
- [ ] Follow-up: the `tests/TLib.py` snapshot/restore workaround from #640 can be simplified or removed.

### Target

Post-1.10 (behavior-adjacent; not for the conservative 1.10 release line). Relates to #640.

</details>

thiell:
Follow the [Logging HOWTO guidance for libraries](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library): add a `NullHandler` to the top-level `ClusterShell` logger.

- Without it, an application embedding the library without configuring logging gets ClusterShell's WARNING+ records raw on stderr via `logging.lastResort` (Python 3.2+), or a "No handlers could be found" message (Python 2.7).
- All library emission sites use `getLogger(__name__)` under the `ClusterShell` namespace, so one handler covers them all.
- No CLI change: `clush -d` debug output is unchanged (records still propagate to the root handler); normal runs stay quiet; gateway file logging unaffected.
- Behavior note: embedders that relied on the implicit `lastResort` output must now configure logging to see ClusterShell messages — the standard opt-in library posture.
- `logging.NullHandler` is available since Python 2.7.

Documented in a new *Library logging* section of the Programming Guide (logger namespace, the `NullHandler` posture, how to enable messages, and a warning about the behavior change), as asked in review.

The second commit covers the one place where `lastResort` was doing real work: a gateway relays its stderr to the root node through the propagation channel, so `Gateway failure:` messages stayed visible even when `CLUSTERSHELL_GW_LOG_DIR` is not writable. With the `NullHandler` alone they go silent, leaving only exit code 1 — so `gateway_main()` now falls back to stderr logging when the log file cannot be opened. Checked by hand with an unwritable log dir; the writable case is unchanged.

Closes #641
See #687

### DIFF
--- a/doc/sphinx/guide/index.rst
+++ b/doc/sphinx/guide/index.rst
@@ -2,11 +2,11 @@ Programming Guide
 =================
 
 This part provides programming information for using ClusterShell in
-Python applications. It is divided into two sections: node sets handling and
-cluster task management, in that order, because managing cluster tasks
-requires some knowledge of how to deal with node sets. Each section also
-describes the conceptual structures of ClusterShell and provides examples of
-how to use them.
+Python applications. It starts with node sets handling, then cluster task
+management, in that order, because managing cluster tasks requires some
+knowledge of how to deal with node sets. These sections also describe the
+conceptual structures of ClusterShell and provide examples of how to use
+them. A separate section explains how the library logs messages.
 
 This part is intended for intermediate and advanced programmers who are
 familiar with Python programming and basic concepts of high-performance
@@ -18,4 +18,5 @@ computing (HPC).
     nodesets
     rangesets
     taskmgnt
+    logging
     examples
--- a/doc/sphinx/guide/logging.rst
+++ b/doc/sphinx/guide/logging.rst
@@ -1,0 +1,64 @@
+.. _library-logging:
+
+Library logging
+===============
+
+.. highlight:: python
+
+Logger namespace
+----------------
+
+ClusterShell uses the standard Python ``logging`` module. Each module logs
+through a logger named after itself, below the top-level ``ClusterShell``
+logger, for example ``ClusterShell.NodeUtils``, ``ClusterShell.Task`` or
+``ClusterShell.Worker.Tree``.
+
+Following the `Logging HOWTO recommendation for libraries
+<https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library>`_,
+ClusterShell never configures logging on behalf of your application: it only
+adds a ``NullHandler`` to its top-level logger. Nothing is logged anywhere
+until your application configures logging.
+
+.. warning:: Since version 1.11, an application that does not configure
+   logging no longer sees ClusterShell warnings and errors on standard error.
+   Such messages were previously printed by the ``logging`` module handler of
+   last resort. Configure logging as shown below to get them back.
+
+.. _library-logging-enabling:
+
+Enabling ClusterShell messages
+------------------------------
+
+The simplest way is to configure logging globally in your application::
+
+    import logging
+
+    logging.basicConfig(level=logging.WARNING)
+
+To enable ClusterShell messages only, and leave the logging configuration of
+the rest of your application untouched, add a handler to the ``ClusterShell``
+logger instead::
+
+    import logging
+
+    handler = logging.StreamHandler()  # stderr
+    handler.setFormatter(logging.Formatter('%(name)s: %(levelname)s: '
+                                           '%(message)s'))
+
+    logger = logging.getLogger('ClusterShell')
+    logger.addHandler(handler)
+    logger.setLevel(logging.WARNING)
+
+The same applies to a single module: use ``ClusterShell.NodeUtils`` to get
+node group resolution messages only.
+
+.. note:: Most ClusterShell log records are ``DEBUG`` ones written to
+   troubleshoot the library itself. They are verbose and contain object
+   representations and engine internals. Use ``WARNING`` and above to only get
+   messages that report an actual problem.
+
+The command-line tools configure logging themselves, so this does not affect
+them: see the ``-d`` option of :ref:`clush <clush-tool>`,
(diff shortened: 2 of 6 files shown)


degremont:
Looks OK, but that probably needs to be explain somewhere in the doc, no?

thiell:
Postponing for now, but keeping this PR open. Maybe we do that for a 1.11 with corresponding documentation indeed.

degremont (approved):

PR #649, opened 2026-07-06
CI: add docs build workflow (sphinx -W)

thiell:
Add a GitHub Actions workflow that builds the Sphinx docs with warnings as errors, so doc breakage fails PRs early instead of post-merge on ReadTheDocs.

- Mirrors RTD: same pins (`doc/sphinx/requirements.txt`) and same Python (3.12) as `.readthedocs.yaml`; RTD already enforces `fail_on_warning: true`.
- Runs on all PRs (no paths filter): autodoc imports the library, so broken docstring markup in `lib/` fails the build too.
- Includes the `merge_group` trigger like tests.yml so the check reports in the merge queue.

After merge, the `sphinx` check (shown as "docs / sphinx" on PRs) can be added to the required status checks to make it blocking.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,21 @@
+name: docs
+
+on: [push, pull_request, merge_group]
+
+permissions:
+  contents: read
+
+jobs:
+  sphinx:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


degremont (approved):
LGTM. Did we have a situation where we merged broken doc changes?

thiell:
I recently fixed old typos in the sphinx doc that could have been avoided by a simple CI check like that, I'm pretty sure.

PR #654, opened 2026-07-08
Documentation: port examples to Python 3

thiell:
Port the documentation examples to Python 3:

- programming guide: `print` statements, `iteritems()`, and bytes message
  buffers (`str()` of a message buffer raises TypeError on Python 3; use
  `msg.decode()` / `buffer.message().decode()`)
- check_nodes.py: same, plus datetime.strptime()-based date parsing to avoid
  a yearless-date DeprecationWarning on Python 3.13+
- pp example: fix `tuple(nodelist)` NameError and reference ppft, the
  Parallel Python fork for Python 3
- NodeSet/Task/Worker docstring examples (rendered in the API docs): same

All examples were tested on Python 3.9 (Rocky 9) and 3.13 (Debian 13):
guide doctests including the node group examples, Task/EventHandler examples
over ssh, check_nodes.py end-to-end, and a sphinx -W build with the CI pins.


README.md
```diff
@@ -98,13 +98,12 @@ Python code (simple example)
 >>> from ClusterShell.NodeSet import NodeSet
 >>> task = task_self()
 >>> task.run("/bin/uname -r", nodes="linux[4-6,32-39]")
-<ClusterShell.Worker.Ssh.WorkerSsh object at 0x20a5e90>
+<ClusterShell.Worker.Ssh.WorkerSsh object at 0x7f45287db910>
 >>> for buf, key in task.iter_buffers():
-...     print NodeSet.fromlist(key), buf
+...     print(NodeSet.fromlist(key), buf.message().decode())
```
degremont, line 103: a pity iter_buffers() is now that complex to use (`buf.message().decode()`), compared to the original intent.

doc/sphinx/guide/examples.rst
```diff
@@ -118,18 +122,18 @@ is usually packaged with ClusterShell::
...
             date_boot = None
             for fmt in ("%Y-%m-%d %H:%M",): # formats with year
                 try:
-                    # datetime.strptime() is Python2.5+, use old method instead
-                    date_boot = datetime(*(time.strptime(bootime, fmt)[0:6]))
+                    date_boot = datetime.strptime(bootime, fmt)
                 except ValueError:
                     pass
-            for fmt in ("%b %d %H:%M",):    # formats without year
+            for fmt in ("%Y %b %d %H:%M",): # formats without year
```
degremont: comment is now confusing

doc/sphinx/guide/examples.rst
```diff
@@ -118,18 +122,18 @@ is usually packaged with ClusterShell::
         def ev_read(self, worker, node, sname, msg):
             """Read event from remote nodes"""
             # this is an example to demonstrate remote result parsing
-            bootime = " ".join(msg.strip().split()[2:])
+            bootime = " ".join(msg.decode().strip().split()[2:])
             date_boot = None
             for fmt in ("%Y-%m-%d %H:%M",): # formats with year
```
degremont: why do we need a `for-loop` in this example?

degremont (commented):
> bytes message buffers (str() of a message buffer raises TypeError on Python 3; use
msg.decode() / buffer.message().decode())

Is there a way to keep that simple? Why `str()` is raising a TypeError now? This is our own object no?

README.md
thiell: This is the behavior since 1.8 to support Python 3. Buffers are `bytes` and need to be properly decoded for printing. Casting to string is not supported because of that. We could add a `MsgTreeElem.decode()` function in 1.11 so the example would become `buf.decode()`, would that work?

doc/sphinx/guide/examples.rst
thiell: Good point, will fix!

doc/sphinx/guide/examples.rst
thiell: Right, it's not ideal. I can replace this example with something like this instead:

```python
    def ev_read(self, worker, node, sname, msg):
        """Read event from remote nodes"""
        # this is an example to demonstrate remote result parsing
        bootime = " ".join(msg.decode().strip().split()[2:])
        # 'who -b' prints the boot date with or without the year
        try:
            date_boot = datetime.strptime(bootime, "%Y-%m-%d %H:%M")
        except ValueError:
            try:
                # no year: prepend the current one before parsing
                # (yearless dates are deprecated as of Python 3.13)
                date_boot = datetime.strptime(
                    "%d %s" % (date.today().year, bootime), "%Y %b %d %H:%M")
            except ValueError:
                date_boot = None
```

README.md
degremont: that maybe a good idea, but i need to refresh my mind why we need this.

I don't think that regular Py3 code needs to call decode() all the time, like when printing the content from a file

degremont: OK. I refreshed my mind. 
We should create a ticket for 1.11 where we implement some helpers.

1. A text=True/encoding= style option, mirroring subprocess and open(). A task- or worker-level default encoding (e.g. Task.set_default('encoding', 'utf-8')) would let iter_buffers() optionally yield decoded str, with a documented error policy (errors='replace'/'surrogateescape'). This keeps bytes as the safe default but offers a one-line text mode.
4. A convenience decode method on the element, e.g. buffer.decode(encoding='utf-8', errors='replace'), so the common case is buffer.decode() instead of buffer.message().decode(). This is a thin wrapper — message() already returns bytes, so MsgTreeElem could just proxy bytes.decode.
6. surrogateescape by default for a decode() helper would make it lossless and never crash on non-UTF-8 bytes — the pragmatic choice for a tool that shells out to arbitrary commands.

degremont (approved):

Issue #485, opened 2022-08-30
Performance on large many-dimensional nodesets


mattaezell:
HPE Cray EX supercomputers use hardware locations (called xnames) that encode up to 5 dimensions.  An example compute node might be `x1000c2s3b0n1`. We don't use xnames for our compute nodes, but our switches, bmcs, chassis controllers, etc do use them.  Some of our local tooling (and clush/cluset) struggle with long lists of xnames, particularly when folding.
```
[root@admin1.borg ~]# time (cluset -e x[2000-2073]c[0-7]s[0-7]b[0-1] | cluset -f)
x[2000-2073]c[0-7]s[0-7]b[0-1]

real    6m38.483s
user    6m38.228s
sys     0m0.088s
```
Any thoughts on how we could improve this? Thanks

thiell:
Thanks for the report @mattaezell.

A quick look (on the master branch) shows that most of the time is spent in `RangeSetND._fold_multivariate_merge()` which does the nD folding:

```
$ cluset -e x[2000-2073]c[0-7]s[0-7]b[0-1] | python3 -m cProfile -s cumulative lib/ClusterShell/CLI/Nodeset.py -f
x[2000-2073]c[0-7]s[0-7]b[0-1]
         1225113449 function calls (1073087573 primitive calls) in 848.288 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    107/1    0.009    0.000  848.388  848.388 {built-in method exec}
        1    0.000    0.000  848.388  848.388 Nodeset.py:27(<module>)
        1    0.001    0.001  848.258  848.258 Nodeset.py:334(main)
        1    0.000    0.000  848.257  848.257 Nodeset.py:155(nodeset)
        1    0.251    0.251  848.255  848.255 Nodeset.py:44(process_stdin)
    18950    0.028    0.000  846.483    0.045 RangeSet.py:884(inner)
152119181/94356   14.296    0.000  845.921    0.009 {built-in method len}
     9473    0.008    0.000  845.920    0.089 RangeSet.py:1147(_fold)
        3    0.000    0.000  845.911  281.970 NodeSet.py:238(__len__)
        3    0.000    0.000  845.911  281.970 RangeSet.py:926(__len__)
        1    0.000    0.000  845.856  845.856 RangeSet.py:1180(_fold_multivariate)
[...]

PR #656, opened 2026-07-09
RangeSetND: speed up multivariate folding

thiell:
Follow-up to #485 — rewrite of the nD folding internals: expand groups unique elements by leading coordinates in one pass (no per-element objects), and merge replaces pairwise vector comparison with hash-based grouping on the non-differing axes.

| shape | before | after |
|---|---|---|
| `x[2000-2073]c[0-7]s[0-7]b[0-1]` from 9472 nodes (#485) | 0.32s | 0.06s |
| 3000 scattered 2D points | 26s | 0.02s |
| 4000 scattered 3D points | 39s | 0.05s |
| two overlapping 300x300 boxes | 11.2s / 237MB | 0.02s / 15MB |
| 100x100 grid from 10000 nodes (already fast) | 0.19s | 0.02s |
| 100k scattered 4D xnames, `cluset -f` end to end | impractical | 6.2s |

Folded results always represent the same node set, but the decomposition of overlapping nD inputs may differ (often fewer vectors); test expectations updated, as previously done in ec14536.

Validation:
- 44k+ differential fuzz cases vs master (element sets, length, padding, disjointness, idempotency): zero failures
- output canonicity verified across build paths and hash seeds
- identical fold output digests on Python 2.7.5, 3.9 and 3.13; test suite green on all three
- no shape found where the new code is slower

Also adds a short "How nD folding works" overview to the docs.

### DIFF
--- a/doc/sphinx/tools/nodeset.rst
+++ b/doc/sphinx/tools/nodeset.rst
@@ -565,6 +565,38 @@ contiguous node set in an expanded manner to separate lines::
     node11 node12 node13 node14 node15 node16 node17 node18 node19
 
 
+How nD folding works
+""""""""""""""""""""
+
+Internally, a multidimensional node set is stored as a list of *range
+vectors*, one range set per axis: ``x[1-2]c[1-4]`` is the vector
+``(1-2; 1-4)`` and represents the Cartesian product of its axes, that is
+2 x 4 = 8 nodes. When node sets are combined, ClusterShell *folds* the
+result in two steps: it first rewrites it as vectors that do not overlap,
+then repeatedly merges any two vectors that are identical on all axes but
+one — the only kind of merge that is always exact (for example,
+``x1c[1-4]`` and ``x2c[1-4]`` merge into ``x[1-2]c[1-4]``). Vectors are
+finally sorted, larger first. The folded output is canonical: it depends
+only on the resulting set of nodes, not on the order in which node sets
+were combined or expressed.
+
+Folding aims for compact output but does not guarantee the smallest
+possible number of vectors, as computing an optimal decomposition is a
+hard combinatorial problem. The exact decomposition of an equivalent node
+set may also vary between ClusterShell releases, so scripts should compare
+node sets with set operations rather than by their string representations.
+For example, to check that two node sets are equal, use ``-X`` (symmetric
+difference) to count the nodes that are in one set but not in the other::
+
+    $ nodeset -c "x1c[1-4],x2c[1-4]" -X "x[1-2]c[1-4]"
+    0
+
+A count of 0 means the two node sets contain exactly the same nodes. With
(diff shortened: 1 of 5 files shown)


thiell:
Refreshed after rebasing onto master `bcc8d83`, which now includes #657 and #659 — the numbers in the description predate both.

**Changes in the branch since the last push**

- the merge phase unions through the public `update()` instead of a raw `set.update()`, so the sorted-element cache added by #657 is always invalidated; the two remaining raw `set` calls only fill freshly created vectors and carry a `# fresh set` marker
- a new test rejects any other raw `set` mutation in `RangeSet.py`
- `test_folding_canonical`: the same 13 nodes built three ways — as two overlapping boxes, as the same boxes in the opposite order, and as 13 individual nodes in scrambled order — must fold to an identical string, so the result depends only on the node set and not on how the caller assembled it
- `test_fold_update_fold` gains a `difference_update()` step: that operation leaves an axis `RangeSet` referenced by more than one vector, so it is where an in-place mutation or a stale cached view would show up first

**Fold, timed in isolation** — CPython 3.13, input rebuilt for every repetition, min of 3, one process per measurement. Both columns come from the same run, so absolute values differ slightly from the table in the description.

| input | master `bcc8d83` | this branch | speedup |
|---|---|---|---|
| `x[2000-2073]c[0-7]s[0-7]b[0-1]` from 9472 nodes (#485) | 351 ms | 81 ms | 4.3x |
| 100x100 grid from 10000 nodes | 193 ms | 24 ms | 7.9x |
| 3000 scattered 2D points | 295 ms | 9.6 ms | 31x |
| 4000 scattered 3D points | 17.6 s | 38 ms | 458x |
| 5000 scattered 4D points | > 300 s | 90 ms | > 3300x |
| two overlapping 300x300 boxes | 6.42 s, 248 MB peak | 18 ms, 15 MB peak | 353x, 16x less memory |
| 3103 xnames of a 60%-populated cabinet | 5.06 s | 38 ms | 134x |

End to end on that last one, `nodeset -f` (`cluset -f`) over the 3103 names: 5.20 s → 0.25 s. Node name parsing is untouched and dominates the cheaper cases, so end-to-end gains are smaller than fold-only gains except where the old fold was the bottleneck.

Validated against master with a local differential harness (not part of this PR): a catalog of generated shapes — 1D to 6D, dense grids, sparse scatter, overlapping and nested boxes, diagonals, mixed padding, autostep, real cluster layouts — plus randomized fuzzing, all confirming that both trees fold to exactly the same node set, that the output is independent of hash seed and build order, and that no shape is measurably slower.

One correction to the description: it says the decomposition of overlapping nD input "often" has fewer vectors; on a larger sample it is closer to even, with roughly as many cases folding to fewer vectors as to more. The node set is identical either way.

degremont (approved):
Impressive. I did not validate the new algo.

OK. Now let's have cluster nodenames with 5 dimensions everywhere! 😁

PR #657, opened 2026-07-09
RangeSet: cache sorted elements between mutations

thiell:
Cache the result of `RangeSet._sorted()` and invalidate it on mutation, instead of re-sorting the inner set on every read. Caching only, no behavior change.

- All reads already funnel through `_sorted()`, so the cache lives in that one method.
- Every mutating method drops the cache with an inline `self._sorted_cache = None` as its first statement — before the mutation runs, so a failing mutator cannot leave a stale view. An earlier revision used a decorator; a Python frame per mutation cost ~200ns against ~25ns for the plain store, which is +65% on a `RangeSet.add()` loop.
- New `pop()` override: it was inherited from `set` and would have bypassed invalidation.
- New unit tests cover invalidation across every mutator, `pop()`, and folding interleaved with updates. A further test derives the mutator list as `set(vars(set)) - set(vars(frozenset))` — those are exactly the in-place mutators — and asserts `RangeSet` overrides each one, so a new `set` mutator in a future Python fails the test instead of silently bypassing the cache.
- No `RangeSetND` change needed here: nD folding only uses public `RangeSet` operations. #656 adds a raw `set.update(gvec[pos], rgvec[pos])` in `_fold_multivariate_merge`; whichever of the two lands second should route that through `update()` so the cache is dropped.

Read times (Python 3.13, 100k-element sets):

| operation | before | after |
|---|---|---|
| `str(rs)` | 139ms | 23ms cached, 133ms first call |
| `list(rs)` | 108ms | 0.8ms |
| `rs[50000]` | 109ms | 0.2us |
| `rs.split(32)` | 3.5s | 124ms first call, 4.2ms cached |
| `NodeSet("node[0-99999]").split(32)` | 3.6s | 7.2ms |
| `str()` of the same NodeSet | 140ms | 25ms |

Per-element mutation costs ~25ns more per call (+7% on a 50k `add()` loop with integers, +21% with strings, unchanged on `add_range()` and the set operators).

Extra memory: ~0.8MB per 100k elements while the cache is live.

Also tested on Python 2.7.5 (full RangeSet/NodeSet suite + randomized differential fuzzing).

### DIFF
--- a/lib/ClusterShell/RangeSet.py
+++ b/lib/ClusterShell/RangeSet.py
@@ -121,12 +121,15 @@ class RangeSet(set):
     """
     _VERSION = 4    # serial version number
 
+    _sorted_cache = None    # built by _sorted(), reset by every mutator
+
     def __init__(self, pattern=None, autostep=None):
         """Initialize RangeSet object.
 
         :param pattern: optional string pattern
         :param autostep: optional autostep threshold
         """
+        self._sorted_cache = None
         set.__init__(self)
 
         if pattern is not None and not isinstance(pattern, basestring):
@@ -308,9 +311,11 @@ class RangeSet(set):
         return (len(elem), elem)
 
     def _sorted(self):
-        """Get sorted list from inner set."""
-        # For mixed padding support, sort by both string length and index
-        return sorted(set.__iter__(self), key=self._sortkey)
+        """Get sorted list from inner set (cached; callers must not mutate)."""
+        if self._sorted_cache is None:
+            # For mixed padding support, sort by both string length and index
+            self._sorted_cache = sorted(set.__iter__(self), key=self._sortkey)
+        return self._sorted_cache
 
     def __iter__(self):
         """Iterate over each element in RangeSet, currently as integers, with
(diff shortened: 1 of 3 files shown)


degremont:
What if a future Python version adds a new `set` mutator?

thiell:
Good point — a new `set` mutator would silently bypass the cache.

That is what `test_set_mutators_are_overridden` guards: the methods `set` has and `frozenset` does not are exactly the in-place mutators, so the test derives that list instead of hardcoding it and checks `RangeSet` overrides each one. A new mutator lands in the list and fails the test with its name, in CI on the new Python. `test_sorted_view_after_mutation` covers the mutators we have today

Also since your review: invalidation is now inline in each mutator instead of the decorator, which cost ~200ns per call against ~25ns for the store.

degremont (approved):
`__init__()` does not invalidate the cache. This is pretty unusual to call it directly, but i don't know, maybe some inheritance or something?

Otherwise, good!

thiell:
Good catch. I flagged that too in an AI session. ClusterShell never reinitializes a live RangeSet but it's so simple to fix it's worth doing so `__init__()` now drops the cache as its first statement like the other mutators. Added test for it too.

degremont (approved):

Issue #337, opened 2017-04-04
Python 3 support


grafoo:
hi!
will python3 be supported in any foreseeable future?

cheers, manuel

thiell:
Hi @grafoo. AFAIK, ClusterShell is mainly used in the HPC world where Python 3 is not available at the default system Python interpreter. The HPC software stack takes time to evolve and thus we also had to ensure compatibility with old Python versions, like Python 2.4 on CentOS 5 and 2.6 on CentOS 6.

But right now, we have a few open tickets related to Python 3 support, as for example:
- #239 is handling the support drop of Python 2.5 and below (first patches landed)
- #316 is handling the migration to argparse

Can you please describe more in detail your use-case of ClusterShell and why Python 3 would be useful in your case? Thanks!

grafoo:
sorry for the delay.
there is no particular use-case for python3.
i'm in the middle of evaluating different frameworks and python3-support was just one item on the pro/con-list.
thanks for the fast response and keep up the good work!
cheers, m

degremont:
For your information we are very closed to have master branch supporting Python 3 for "nodeset" like features. Support for clush & co will probably be more difficult and will require more time.

degremont:
I'm reopening this ticket to follow Python 3 support effort. There is a couple of patch in the pipe for that.

volans-:
@thiell It's very nice to see some effort in this direction, I'm looking forward to the Python 3 support too.

The main reason being that I've developed a tool (usable both via CLI and as a Python library) that uses ClusterShell as a library and now I'm starting to develop additional softwares that uses this tool as a library.
ClusterShell is the only dependency non-Python 3 ready and this means that the tool itself had to be written in Python 2 and as a consequence also any software that uses it as a library. This will generate quite a bit of technical debt and being able to start already in Python 3 would really help.

I'm also available to help, so if you have any specific subtasks for the Python 3 support feel free to loop me in and I'll see what I can do.

thiell:
@volans- thanks for your feedback, much appreciated! As you have probably noticed, things are moving forward.  If you want to contribute, feel free to submit a PR at any time, and as for code review, we still work on gerrithub so feel free to review any patch there. Also, we will welcome Python 3 testers as soon as this work is completed.

volans-:
@thiell Thanks, will do!
Although I'll have to do it from a different account, my GitHub name is not a valid Gerrit name so I can't login on GerritHub with this one 😞

[...]

PR #658, opened 2026-07-09
Fix gateway message pickle protocol and empty RangeSet pickle

thiell:
Tree mode pickles gateway messages with the default pickle protocol. Python 3.14 bumped the default to protocol 5, unreadable by Python < 3.8, so a tree mixing a py3.14+ node with a py3.6/3.7 node (e.g. EL8) breaks in either direction (commands down, results up).

- Pin the gateway message pickle protocol to `min(pickle.HIGHEST_PROTOCOL, 4)` — readable since Python 3.4 (protocol 2 on Python 2, up from 0). This only changes what a node writes; no supported pairing regresses and mixed py2/py3 trees remain unsupported as before.
- Catch `ValueError` in `Message.data_decode()` (a too-new protocol raises `ValueError`, not `UnpicklingError`, and escaped as a raw traceback) and include the cause in `MessageProcessingError`.
- Fix unpickling of empty `RangeSet` objects: `__reduce__()` used `str(self)`, but `RangeSet("")` raises by design, so empty sets (also embedded in `NodeSetBase`/unfolded `RangeSetND` pickles) could be pickled but never unpickled.

Tested on Python 2.7, 3.13 and 3.14 (Tree/Gateway/RangeSet/NodeSet test modules).

### DIFF
--- a/lib/ClusterShell/Communication.py
+++ b/lib/ClusterShell/Communication.py
@@ -47,6 +47,8 @@ try:
 except ImportError:  # Python 2 compat
     import cPickle
 
+from pickle import HIGHEST_PROTOCOL
+
 import base64
 import binascii
 import logging
@@ -74,6 +76,10 @@ ENCODING = 'utf-8'
 
 # See Message.data_encode()
 DEFAULT_B64_LINE_LENGTH = 65536
+
+# Gateway messages may cross Python versions within a tree: cap the pickle
+# protocol at 4, readable since Python 3.4 (Python 2 writes protocol 2)
+GW_PICKLE_PROTOCOL = min(HIGHEST_PROTOCOL, 4)
 
 
 class MessageProcessingError(Exception):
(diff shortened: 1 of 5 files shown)


lib/ClusterShell/Communication.py
```diff
@@ -75,6 +77,10 @@
 # See Message.data_encode()
 DEFAULT_B64_LINE_LENGTH = 65536
 
+# Gateway messages may cross Python versions within a tree: cap the pickle
+# protocol at 4, readable since Python 3.4 (Python 2 writes protocol 2)
+GW_PICKLE_PROTOCOL = min(HIGHEST_PROTOCOL, 4)
```
degremont, line 82: if we are supporting py2 in 1.10, should we cap it to 2 instead?

thiell: Mixed py2/py3 trees are already unsupported regardless of the pickle protocol: the docs require "the same major version of Python on the gateways and the root node" (https://clustershell.readthedocs.io/en/latest/tools/clush.html#remote-python-executable, since 1.8.1), and the root node enforces it by launching gateways with the same executable name, `python` vs `python3` (0d83413, #337).

So capping at 2 would only penalize Python 3 trees (protocol 2 is much less efficient there) without enabling anything new. With `min(HIGHEST_PROTOCOL, 4)`, Python 2 trees already use protocol 2.

degremont (approved):

PR #659, opened 2026-07-09
RangeSet: compute padding without sorting

thiell:
Perf follow-up for 1.11 (independent of #657, based on master):
- The `padding` getter sorted the whole set and int()-converted every element just to find the largest digit length; it now scans the raw inner set and tracks the max.
- `rs.padding` (Python 3.13, 100k-element set): 117ms -> 10ms. Pickling benefits indirectly since `__reduce__` reads `self.padding`: `pickle.dumps` on the same set: 257ms -> 136ms.
- No behavior change; getter test cases added.

### DIFF
--- a/tests/RangeSetTest.py
+++ b/tests/RangeSetTest.py
@@ -1266,6 +1266,16 @@ class RangeSetTest(unittest.TestCase):
         self.assertEqual(r0.padding, None)
         self.assertEqual(str(r0), "0-10,15-20")
 
+    def test_padding_property_getter(self):
+        # largest padding wins
+        self.assertEqual(RangeSet("5,05,005").padding, 3)
+        self.assertEqual(RangeSet("00-99,000").padding, 3)
+        # longer unpadded elements do not mask a padded one
+        self.assertEqual(RangeSet("05,100").padding, 2)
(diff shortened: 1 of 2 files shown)


tests/RangeSetTest.py
```diff
@@ -1266,6 +1266,15 @@ def test_padding_property_compat(self):
         self.assertEqual(r0.padding, None)
         self.assertEqual(str(r0), "0-10,15-20")
 
+    def test_padding_property_getter(self):
+        # largest padding wins regardless of element order
+        self.assertEqual(RangeSet("5,05,005").padding, 3)
+        # longer unpadded elements do not mask a padded one
+        self.assertEqual(RangeSet("05,100").padding, 2)
+        self.assertEqual(RangeSet("1-10").padding, None)
+        self.assertEqual(RangeSet("-10--5,05").padding, 2)
```
degremont, line 1276: Love it. Keeping it for future ClusterShell Trivia in a trendy beer bar in West Village :D

degremont (approved):
Could go in a `1.10.2` or `1.11`. This is a small enough change.

We could decide later if we cut a 1.10.2 or 1.11.

PR #665, opened 2026-07-12
completion: add tests for the bash completion scripts

thiell:
Functional tests for the `bash_completion.d` scripts: run `_clush`/`_cluset` in a child bash sourcing the real bash-completion library and check `COMPREPLY`.

- regression guards for the no-arg option bugs fixed in #619 and #620
- glued short option completion (`-wNODE`, #586), `--color`/`-R` value sets, command and file completion, and hermetic node/group completion using a `cluset` stub on a prepended `PATH`
- a `bash -n` syntax gate that needs only bash
- install bash-completion in CI so the functional tests run there (they skip gracefully where it is missing)

Each functional test was mutation-checked: re-introducing the old bugs, or breaking the specific script behavior a test covers, makes exactly that test fail.

See #619, #620.


degremont (approved):
This is an important enough feature to add a test for it, will ensure we do not break that too easily.

PR #666, opened 2026-07-13
doc: explain DNS domain suffix stripping for Ansible bindings

thiell:
When `ansible-inventory` returns fully qualified hostnames, `nodeset`/`cluset`/`clush` output is more verbose than needed on clusters where short names resolve.

- `config.rst`: document how to strip a DNS domain suffix in the jq filters, in the same spirit as the existing `ansible_host` adaptation note
- `ansible.conf.example`: add a pointer comment (upcalls unchanged)


degremont (approved):
LGTM, but the example config files are now super official and their docs are in the official doc. Not what i was expecting originally, but well, good enough. 
Just wondering where is the right place to add doc for that...

PR #677, opened 2026-07-17
tests: fix race in testThreadTaskUnhandledException

thiell:
Fixes a flaky test seen on master (Python 3.14 job, `AssertionError: b'' != b'raisefoobar'`): https://github.com/clustershell/clustershell/actions/runs/29529565499/job/87726353955

- after an unhandled exception in a Task thread, `_thread_start()` runs `excepthook` then `_terminate(kill=True)`, which resets task buffers
- `join()` returns as soon as `_resume()`'s `finally` block notifies, so reading `task.key_buffer()` afterwards races with that reset; the test usually won only because the default excepthook's stderr traceback slows the task thread down (reproduced deterministically by silencing excepthook)
- check the message captured by the event handler instead, which still verifies the line was delivered to `ev_read()` before the exception was raised

### DIFF
--- a/tests/TaskThreadJoinTest.py
+++ b/tests/TaskThreadJoinTest.py
@@ -108,15 +108,19 @@ class TaskThreadJoinTest(unittest.TestCase):
         class TestUnhandledException(Exception):
             """test exception"""
         class RaiseOnRead(EventHandler):
+            def __init__(self):
+                self.msg = None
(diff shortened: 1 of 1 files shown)


degremont (approved):
Could go to a `1.10.2` or `1.11` as this is a bugfix. Could be merged right away.

PR #678, opened 2026-07-17
Tree: support optional route weights in topology.conf (#374)

### LINKED ISSUE #374: Gateway: add hop/weight parameter to routes
Related to #260. Would be great to have a weighted routes, to either spread the load accordingly, or even able to set up failback routes in topology.conf (eg. a weight of 0 would be used only if that's the only route left).

thiell:
Implement route weights for tree mode, as suggested in #374.

An optional `[weights]` section in `topology.conf` defines relative route weights for gateway nodes (default weight: 1):

```ini
[weights]
gw1: 3
gw2: 2
gwspare: 0
```

- Each target node is dispatched to the gateway with the least number of current connections relative to its weight, so a gateway with weight 3 handles about 3x more targets than a gateway with weight 1. No behavior change when `[weights]` is absent.
- A weight of 0 defines a standby gateway, only used when no other gateway is available for the route, e.g. after all regular gateways were marked unreachable.
- Example: with the weights above and 4 gateways serving `node[001-120]`, dispatch yields 60/40/20/0 targets; once `gw[1-3]` are marked unreachable, `gwspare` takes all 120.
- Weight entries use regular node set syntax (including node groups) and apply to each node of the set individually. Entries are applied in file order (last one wins on overlapping node sets); an entry that does not match any router node is ignored, with a warning when it has no effect at all.
- Weights are carried by the propagated topology tree, so gateways apply them at deeper levels of the tree; topology trees from older versions (without weights) still work.
- Older ClusterShell versions ignore the new section, so a `topology.conf` using `[weights]` stays compatible with them.
- Add `PropagationTreeRouter` unit tests (new `tests/TreeRouterTest.py`), parser tests, and documentation (config.rst, clush.rst tree mode section, topology.conf.example).

Closes #374

### DIFF
--- a/conf/topology.conf.example
+++ b/conf/topology.conf.example
@@ -10,3 +10,11 @@
 rio0: rio[10-13]
 rio[10-11]: rio[100-240]
 rio[12-13]: rio[300-440]
+
+# Optional relative weights for gateway nodes (default weight: 1).
+# A gateway with weight 2 is selected for twice as many target nodes as a
+# gateway with weight 1. A weight of 0 defines a standby gateway, only used
+# when no other gateway is available for a route.
+#[weights]
+#rio10: 2
+#rio13: 0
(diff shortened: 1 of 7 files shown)


thiell:
Superseded by #679

PR #688, opened 2026-07-30
NodeUtils: load groups.conf.d and groups.d from all config dirs (#685)

### LINKED ISSUE #685: User-level `groups.conf.d` is silently ignored without a `groups.conf`
Dropping a group-source file into `~/.config/clustershell/groups.conf.d/` has no effect unless a `groups.conf` **also exists in `~/.config/clustershell/`**. The source is silently ignored, the only symptom is `Unknown group source` when you try to use it.

The file does not need any particular content: an **empty** `groups.conf` is enough. Nothing in the documentation or in the shipped `groups.conf` comments mentions it.

Two options here:
* Resolve `$CFGDIR` per candidate directory rather than only for the last parsed `groups.conf`, so every config directory in the search path contributes its own `groups.conf.d`. This is a behavior change but I think it's closer to what users are expecting... Maybe ok to introduce in 1.11?
* Emit a warning when a `groups.conf.d` or `groups.d` directory exists in a config search path but is skipped because that directory has no `groups.conf`. Add a diag info to the user (better than today) but not ideal IMO.

thiell:
`$CFGDIR` in *confdir* and *autodir* was substituted with a single directory: the one holding the highest-priority `groups.conf` that actually exists. Dropping a file into `~/.config/clustershell/groups.conf.d/` without also creating a user `groups.conf` did nothing. `$CFGDIR` is now expanded once per directory of the configuration file search path — including directories that have no `groups.conf`, which is the whole bug.

Since a user drop-in can now meet a same-named system drop-in, config-defined group sources gained override semantics:

- same source name in several scanned directories: the definition scanned last (highest priority) wins; *autodir* sources win over *confdir* ones, and `groups.conf` sections win over all drop-ins
- same source name within one directory: still a fatal error
- public `add_source()` unchanged; overrides are traced with `nodeset -d` / `cluset -d`

For the release notes: previously inert `.conf`/`.yaml` files under a user config dir now go live, and the empty-`groups.conf` workaround is no longer needed (it also moved upcall `$CFGDIR`, silently breaking the shipped `[local]` source).

Docs updated on all three surfaces (`conf/groups.conf`, `groups.conf.txt` + man regen, `config.rst`), including a corrected description of upcall `$CFGDIR` (per-source, set at definition site).

7 new tests (5 red before the fix). Verified beyond CI: layered `XDG_CONFIG_HOME` + `CLUSTERSHELL_CFGDIR` sandbox with `nodeset -d`, and byte-identical `nodeset -LL` output vs master on a real `/etc/clustershell` setup.

Follow-up: `clush.conf.d` has the same single-dir `$CFGDIR` wart (`CLI/Config.py`) — separate issue and PR.

Closes #685


degremont:
Looks overall good, but there are few issue remaining:

* an admin who ships `/etc/clustershell/groups.d/slurm.yaml` cannot be overridden by a user's `~/.config/clustershell/groups.conf.d/slurm.conf`.  Interleaving would fix it: for each config directory in ascending priority, scan that directory's confdir entries and then its autodir entries.
```
DEBUG group source 'slurm' in /tmp/.../etc/groups.d overrides definition in /tmp/.../user/groups.conf.d
P1 user-confdir vs system-autodir -> grouplist: ['sysgroup']
```
* Not super fan of "and groups.conf sections win over all drop-ins". My understanding was rather than drop-in files will have higher priority. The side-effect is that a user could not overload a source defined in `/etc/clustershell/groups.conf`. 

Order by config directory along the search path, and use the file kind only as a tie-break within a directory.

Also, not sure about what we should do, but CLUSTERSHELL_CFGDIR is no longer confines drop-in scanning, with CLUSTERSHELL_CFGDIR pointing at a site config, $XDG_CONFIG_HOME/clustershell/groups.conf.d is now scanned as well.

thiell:
Great points, I will rework this soon. Re: `CLUSTERSHELL_CFGDIR`, I am thinking making sure we have the following:

When `CLUSTERSHELL_CFGDIR` is set, `config_paths()` drops the `/etc/clustershell` entry; everything else (position, XDG, `~/.local`, venv) unchanged.

It was introduced in 1.9 as an "alternate location for global configuration files", where users are expected to keep their personal `~/.config` layer working on top. This was the original idea, not to confine the config.

thiell:
Reworked as suggested, thanks — now 4 commits:

1. Directory-major precedence: config directories are scanned in ascending priority; within a directory, `groups.conf` sections first, then confdir, then autodir; the last definition loaded wins. Your P1 case now resolves to the user's `groups.conf.d/slurm.conf`, a user drop-in overrides a source defined in `/etc/clustershell/groups.conf`, and a section in the user's own `groups.conf` still beats `/etc` drop-ins. Sections are attributed to the directory of the `groups.conf` defining them, so the single merged ConfigParser (option-level merging) is kept. Upcall `$CFGDIR` is now the directory of the defining file, and the shipped `groups.conf` relies on `$CFGDIR` alone — the literal `/etc/clustershell/...` entries were redundant and would have defeated your third point.
2. `clush.conf`'s confdir had the same single-directory `$CFGDIR` wart, which the search path change below would have turned into a real regression for site `clush.conf.d` drop-ins — same fix, sharing the expansion helper.
3. `CLUSTERSHELL_CFGDIR`: I looked at the history and my comment above was only half right. In 1.9 the variable replaced `/etc/clustershell` in its slot, below the per-user files; 1.9.1 (#510) moved it to the very top so that a site config beats the pip-installed defaults in a venv, which also put it above the user's own config, and kept reading `/etc/clustershell`. `config_paths()` now drops `/etc/clustershell` when it is set and places it above the pip layers but below `$XDG_CONFIG_HOME` — so `/etc` is not read, users can still override the site config as they override `/etc`, and the user layer being scanned for drop-ins is the point of #685. Release-note material for all of this is in the PR description.
4. While there, `config.rst` no longer lists `~/.clush.conf` among the files read (not read since 1.9, as the man pages already say).

Issue #687, opened 2026-07-30
Surface useful library warnings in the CLI tools (tracking issue)


thiell:
Today the CLI tools either silence library logging (clush sets the root logger to CRITICAL in normal runs), only configure it with `-d` (nodeset/cluset), or not at all (clubak). So when the library detects something a sysadmin should know about, it's invisible unless you run `-d` and take the full debug firehose. Turning everything on isn't an option either: most library log messages are developer-oriented (object reprs, engine internals) and would look scary in production output.

Plan: standard [library/application logging split](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library), with selective surfacing:

- the library never touches global logging, logs under `ClusterShell.*`, and gets a top-level `NullHandler` (#641) so embedders are unaffected
- the CLI tools attach small stderr handlers to specific `ClusterShell.<module>` loggers only — an allowlist of channels whose messages are written for sysadmins; everything else stays as quiet as today
- a logger joins the allowlist only after its warning messages are reviewed for production quality: name what's affected, the cause, the action taken — no reprs

`clush -d` output and exit codes don't change; the only visible difference is a few new deliberate warning lines on stderr.

Work items (one PR each, linked with `See`):

- [ ] #684: skip-and-warn on illegal group characters — first scoped handler (`ClusterShell.NodeUtils`) in clush, nodeset/cluset and clubak
- [ ] #641 (PR #647): `NullHandler` on the top-level `ClusterShell` logger
- [ ] replace root `logging.basicConfig()` in the CLIs with a shared helper scoped to the `ClusterShell` namespace (keeping `clush -d` output as-is)
- [ ] tree mode: review `Propagation`/`Worker.Tree` warnings and surface gateway failures/rerouting at default verbosity
- [ ] maybe later: scoped debug controls (e.g. tree/communication only)

PR #689, opened 2026-07-30
NodeUtils: warn and skip group names with illegal characters (#684)

### LINKED ISSUE #684: Filter illegal group character errors from external group sources
External group sources can return illegal group characters in group names. It's sometimes hard to predict and filter in advance in the upcalls. I am wondering if we should turn those into warnings instead or erroring:
```
$ cluset -L example-xx1r1c0101
@source1:group1
@source2:group2
@source3:group3
cluset: Illegal group character: "&"
$ echo $?
1
```

In addition, the current error message does not display the bad group, unless `-d` is used, in that case we see:
```
Xabc===&===Ycdf
cluset: Illegal group character: "&"
```

thiell:
Group names returned by group sources (upcalls or group files) that contain illegal characters (`@,!&^*`) are now skipped with a warning instead of aborting the whole operation.

- nodeset/cluset `-l`/`-L`/`-r`, `@source:*` expansion, clush and clubak keep working with the remaining valid groups when an external source returns a bogus entry (exit 0 instead of 1)
- the warning names the offending group and its source, e.g. `cluset: WARNING: ignoring group "Xabc===&===Ycdf" from source "source3": illegal character(s) "&"`, deduplicated per (source, group) as the reverse upcall runs once per node
- example with clush, where a group source returns a bogus entry:
  ```console
  $ clush -bw @* uname
  clush: WARNING: ignoring group "bad&name" from source "local": illegal character(s) "&"
  ---------------
  n[1-2] (2)
  ---------------
  Linux
  ```
- `GroupResolverIllegalCharError` is no longer raised by ClusterShell; the class and its CLI error handling are kept for compatibility, as third-party `GroupResolver` subclasses may still raise it
- the four CLI tools print these warnings through a handler scoped to the `ClusterShell.NodeUtils` logger only, so all other library logging is unchanged (clush stays quiet, `-d` output is untouched)
- tests: `CLI_main()` now restores the `ClusterShell.NodeUtils` logger between in-process CLI runs, like it already did for the root logger
- doc: `groups.conf(5)` and the Sphinx configuration page document the illegal characters and the new behavior

Closes #684. See #687.


degremont:
Not found of this one. This is now a logger warning anywhere in the code when a group resolve with an illegal warning? Whereas the original intent was to be more tolerant with `nodeset -ll` ?

And this being a warning makes that hard to catch programmatically... 

Do not make that an universal behavior. What prevent sanitizing the "list" upcall from the group sources? We are already doing that in slurm.conf example. This is a simple `tr -d ...`.

thiell:
Maybe the logger thing is not the best avenue here, but the problem is that today:
* `nodeset -ll` is NOT tolerant, on the contrary, it aborts when such bogus group shows up, and with dynamic groups, these can show up randomly
* yes, you can manually filter the output on the upcalls, but if you have tens of even hundreds of group sources, you have to do it for all of them, despite the fact that we know exactly the characters that are not supported (basically the arithmetic operators `&^!,` and reserved group prefix `@` ... this makes sense to have them filtered by default I think, so that they are skipped, but something needs to be reported to the sysadmin (in that PR logger is used)

Re: logger, I opened #687 to improve library logging

degremont:
Well, still not fond of this. Should we implement #687 first? as it will change most of what this patch is doing.
Ignoring and reporting on stderr is ok for human command, but not nice for scripts.
Also, what about `clush -w @*`? It will start ignoring nodes instead of reporting error.
Wondering if we should not record the skip records and report them instead, that will make that inspectable from the library too (`Warning: 2 group(s) ignored from source "slurm" (illegal characters))`). 

This is also not inline with `-q` flags, it does not support it.

thiell:
Reworked as you suggested:

* skipped group names are recorded per source and exposed via `GroupResolver.ignored_groups()` — inspectable from the library
* the tools report them with one line per source on stderr, using their existing verbosity mechanisms, so `-q` mutes it:

      Warning: ignoring group names with illegal characters from source "local": "bad&grp", "x^y"

  (the line names the groups — the original complaint in #684)
* clush prints it after target resolution, before running the command
* `clush -w @*`: with an `all:` upcall, the upcall returns nodes, not group names — nothing to check, before or after this patch; without one, the `list`+`map` fallback now reports the warning instead of aborting
* the per-tool logger handler is gone; the library still logs each skip once (`ClusterShell.NodeUtils`). #687 remains the way to surface library logging in the CLI tools, but it is not required for this patch.

lib/ClusterShell/CLI/Nodeset.py
degremont, line 273: you are calling print_ignored_group_warnings() before each return, except here. if `xset` is set, `--completion` will print warnings, but not if not set.

degremont, line 368: not super elegant, but calling print_ignored_group_warnings() here cover all cases...

degremont (commented):
Much better like this. Still few minor comments.
Also, it could be good to add a way to clear the ignored_groups() in UpcallGroupSource.clear_cache(), that will give caller, at least a way to clear it.

lib/ClusterShell/CLI/Nodeset.py
thiell: Reworked. Now both `--completion` cases print the warning by default, but I added `--quiet` to the bash completion scripts to keep it clean, but let me know what you think.

thiell: Ok, but because parsing of options (eg. for `-q`) happens inside `nodeset()`,  I moved the code into a new `run_nodeset()` so that `nodeset()` prints the warnings once after it returns.

degremont:
Almost good, one last thing:

make `--completion` enables `--quiet` automatically. This is mostly a machine-facing argument, and `--completion` is already declared with `optparse.SUPPRESS_HELP` and commented "special hidden command for bash completion scripts"

```
    if options.completion:
        options.quiet = True
```

thiell:
Makes sense, thanks!

degremont (approved):

PR #691, opened 2026-07-31
Task: reset quit flag when a task is resumed

thiell:
`Task.abort()` sets `_quit` and nothing ever cleared it, so a task reused after an abort stayed flagged as quitting for the rest of its life.

Clear it in `_resume()`, which runs on both the `task_self()` and threaded paths, plus a regression test.


lib/ClusterShell/Task.py
```diff
@@ -796,6 +796,8 @@ def _resume(self):
         assert self.thread == threading.current_thread()
         try:
             try:
+                # abort() leaves _quit set: clear it as the task runs again
```
degremont: no need to put a comment every time the code change :)

tests/TaskLocalMixin.py
```diff
@@ -560,6 +560,23 @@ def testTaskAbortSelf(self):
         task.abort()
         self.assertEqual(task, task_self())
 
+    def testTaskResumeAfterAbortSelf(self):
+        task = task_self()
+
+        worker = task.shell("/bin/echo before abort")
+        task.resume()
```
degremont: `task.run()` is prefered.

task.resume(), especially on first run is deprecated

degremont (approved):
One remark about AI putting too many comments, otherwise, looks OK

PR #692, opened 2026-07-31
RangeSetND: fix exponential splitting in difference_update()

thiell:
Removing a few scattered nodes from a large multi-dimensional node set could make ClusterShell hang and eat memory: subtracting just 8 nodes from a folded 100x100x100 RangeSetND never finished, and around 100 nodes it peaked over 800 MB.

The cause is in how `difference_update()` splits vectors. Each subtraction can split a vector into up to dim() pieces, and those pieces overlapped each other — so every subtracted vector could re-split all the pieces of the previous one, multiplying the work at each step. For the mathematicians in the room: the box difference V \ W was decomposed as the union over axes i of V1 x ... x (Vi \ Wi) x ... x Vd, which is correct but not disjoint; narrowing the already-processed axes to their intersection Vi∩Wi makes the pieces pairwise disjoint. A subtracted vector now only touches the few pieces it actually overlaps, and the piece count grows linearly instead of exponentially.

Results: the 8-node case above drops to ~1 ms. Subtracting 5 scattered nodes (the largest case that finished before the fix) goes from 2.1 s to 0.16 s including re-folding, and a 256-node rack minus 3 nodes is 2.4x faster. Cases that were already fast are unchanged.

This also fixes or speeds up everything built on `difference_update()`: `NodeSet` difference (`-x`/`-X` on multi-dimensional node sets), `issuperset()`, `issubset()`, `==`, `in`, and strict removal.

Results are unchanged node-for-node: verified with a 15,000-case differential fuzz against a pure-Python oracle, the full test suites, and a py2.7/EL7 run. One cosmetic note: sets using autostep that lost their step notation after `copy()` + difference (a pre-existing container-autostep quirk) can now keep it, since the disjoint vectors more often take the folding path that preserves axis autostep.

### DIFF
--- a/lib/ClusterShell/RangeSet.py
+++ b/lib/ClusterShell/RangeSet.py
@@ -1463,16 +1463,22 @@ class RangeSetND(object):
                 while len(procvx1) > 0: # refine diff for each resulting vector
                     rgproc1 = procvx1.pop(0)
                     tmpvx = []
+                    # carved axes are narrowed to the intersection so that
+                    # generated vectors are pairwise disjoint
+                    carve = list(rgproc1)
                     for pos, (rg1, rg2) in enumerate(zip(rgproc1, rgvec2)):
-                        if rg1 == rg2 or rg1 < rg2: # issubset
+                        if rg1.issubset(rg2):
                             pass
-                        elif rg1 & rg2:             # intersect
-                            tmpvec = list(rgproc1)
-                            tmpvec[pos] = rg1.difference(rg2)
-                            tmpvx.append(tmpvec)
-                        else:                       # disjoint
-                            tmpvx = [ rgproc1 ]     # reset previous work
-                            break
+                        else:
+                            inter = rg1 & rg2
+                            if inter:               # intersect
+                                tmpvec = list(carve)
+                                tmpvec[pos] = rg1.difference(rg2)
+                                tmpvx.append(tmpvec)
+                                carve[pos] = inter
+                            else:                   # disjoint
+                                tmpvx = [ rgproc1 ] # reset previous work
+                                break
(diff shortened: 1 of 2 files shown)


thiell:
Benchmark details (py3.13, idle host, 7 alternating rounds, min per case; the control case runs identical code in both trees and bounds the noise at about ±1.3%). Before = master 94a06fc, after = this PR:

```
scenario                                       before      after   speedup
control (identical code, noise floor)         66.5 us    67.4 us      1.0x
256-node rack (3D) - 3 scattered nodes        74.5 us    31.5 us      2.4x
rack - one chassis (box - box)                 6.8 us     6.6 us      1.0x
NodeSet 2D 1024 nodes - 2 nodes                255 us     226 us      1.1x
1M box - half (box - box)                     14.8 us    16.0 us     0.93x
1M box - 5 scattered, incl. re-fold            2.09 s     161 ms       13x
1M box - 10 scattered, incl. re-fold           >30 s      159 ms       inf
1M box - 100 scattered, diff only              >30 s     16.6 ms       inf
1M box - 100 scattered, incl. re-fold          >30 s      191 ms       inf
1M box - 1000 scattered, incl. re-fold         >30 s      1.22 s       inf
5D xname box (65k) - 20 nodes, incl. re-fold   >30 s      190 ms       inf
NodeSet 5D 65k - 20 nodes, -x + str            >30 s      156 ms       inf
issuperset() vs a set with 100 holes           >30 s      274 us       inf
```

The "incl. re-fold" rows include re-folding the result, i.e. what `nodeset`/`cluset -x` pays end to end; after the fix those are fold-dominated. Where the old code was already fast (box - box, small 2D), the change is parity within the noise band.

Issue #694, opened 2026-08-04
clush --hostfile does not ignore '#' comments


thiell:
`clush --hostfile` does not handle `#` comments, unlike `cluset`/`nodeset` reading from stdin, which strips them (including inline) since 2009. With a host file like:

```
# managed by ansible
node[1-2]
node3 # rack 12
```

- clush tries to connect to `#`, `managed`, `by`, `ansible`, `rack` and `12` as real targets
- a comment like `# racks [1-3] are down` expands to bogus hosts `1`, `2`, `3`
- a comment with an unbalanced bracket (e.g. `# broken node[5`) aborts clush with a parse error before anything runs

`--hostfile` was added for pdsh/dsh compatibility (#235), and pdsh strips `#` comments in WCOLL files. So do pssh, OpenMPI `mpirun --hostfile`, GNU parallel `--sshloginfile` and Ansible inventories. `--hostfile` should do the same, with the same inline-strip semantics as `cluset` stdin and pdsh.

Related edge found while comparing: the comment-stripping code in `CLI/Nodeset.py` (`line[0:line.find('#')]`) drops the last character when the final stdin line has no trailing newline, e.g. `printf 'node5' | cluset -f -` returns `node`. Both spots should use a safe idiom like `line.split('#', 1)[0]`.

PR #696, opened 2026-08-04
cluset: fix parsing of last stdin line without newline

thiell:
`cluset -f -` silently drops the last character when the final stdin line has no trailing newline: `printf 'node5' | cluset -f -` returns `node`, and `printf 'node[1-2]' | cluset -f -` fails with a parse error.

Cause: `line[0:line.find('#')]` in `process_stdin()` evaluates to `line[0:-1]` when a line contains no `#`, which relied on the trailing newline to be harmless. Use `line.split('#', 1)[0]` instead — same comment-stripping semantics, newline-independent.

Adds stdin tests for an unterminated last line and for `#` comment handling (previously untested).

See #694


degremont (approved):
Why do we even support `#` as comment?

thiell:
Mostly for compat with other tools (see also #698)

PR #697, opened 2026-08-04
EngineClient: fix data loss on CR-separated output

thiell:
`EngineClient._readlines()` splits each read chunk with `bytes.splitlines()`, which treats bare CR as a line break, but the loop assumes only the last fragment can be a partial line: every CR-ended fragment overwrites `rbuf`, so all but the last are silently dropped and the survivor is delivered out of order at EOF. What survives depends on where `os.read()` happens to chunk the stream.

Reproduced with progress-bar-style output (e.g. `dd status=progress`, or tools run with a forced tty):

    printf 'one\rtwo\rdone\nlast\n'
    before: [b'done', b'last', b'two\r']    <- 'one' lost, order scrambled
    after:  [b'one\rtwo\rdone', b'last']

With the pdsh worker this was not only lossy: a CR fragment lacks the `nodename: ` prefix, so `_parse_line()` raised `ValueError` out of the runloop, or attributed one node's output to the next node named in the line.

Fix: split on LF only and trim the CR of a CRLF, so bare CR bytes are preserved verbatim as line data.

Notes:
- CRLF handling is unchanged, including a CRLF split across two reads (covered by tests).
- A stream emitting CR with no LF at all now accumulates in `rbuf` until a newline or EOF, like any other LF-less stream; previously such data was simply lost. Sustained LF-less output pays quadratic `rbuf` re-copies (32 MiB of CR-only output ≈ 8 s, delivered as one message); real progress bars are far smaller, and a follow-up will make the accumulation linear.
- New unit tests cover chunk-boundary buffering (`tests/EngineClientTest.py`), plus end-to-end CR/CRLF tests in `TaskLocalMixin`.

Benchmark of `_readlines()` in isolation (µs/call, one 64 KiB chunk per call, best of 5×3000) — no scenario regressed:

| scenario (64 KiB chunk) | before | after |
|---|---|---|
| 80-byte lines, LF | 147.5 µs | 77.7 µs |
| 80-byte lines, CRLF | 146.2 µs | 108.0 µs |
| 8-byte lines, LF | 1112 µs | 583 µs |
| 80-byte lines, bare CR | 86.1 µs (drops data) | 16.9 µs |
| no newline | 33.1 µs | 16.9 µs |


lib/ClusterShell/Worker/EngineClient.py
```diff
@@ -397,19 +397,15 @@ def _readlines(self, sname):
...
-        rfile.rbuf = bytes()
-        for line in lines:
-            if line.endswith(b'\n'):
-                if line.endswith(b'\r\n'):
-                    yield line[:-2] # trim CRLF
-                else:
-                    # trim LF
-                    yield line[:-1] # trim LF
+        # split on LF only: bare CR is data (eg. progress bars), not a break
+        lines = buf.split(b'\n')
```
degremont, line 400: is that fine when running on system where end of line is not `\n` like Windows or old MacOS ?

thiell: Yep it is fine, a \r immediately followed by \n is removed, exactly as before; a \r anywhere else is now data, which is used for progress bar so this is the biggest risk here, but I think this change makes it more accurate.

Behavior change:

| remote output | before (`splitlines()`) | after (`split(b'\n')` + trim CR of CRLF) |
|---|---|---|
  | `a\nb\n` (POSIX) | `a`, `b` | `a`, `b` |
  | `a\r\nb\r\n` (CRLF, e.g. Windows over ssh) | `a`, `b` | `a`, `b` — the CR before each LF is still stripped |
  | `one\rtwo\rdone\nlast\n` (bare CR, progress bar) | `done`, `last`, then `two\r` at EOF :warning: `one` lost | `one\rtwo\rdone`, `last` — bare CR kept as data |
  | `a\rb\rc\r` (CR-only, no LF at all) | `c\r` at EOF :warning: `a`, `b` lost | `a\rb\rc\r` at EOF (one line) |

degremont (approved):

PR #703, opened 2026-08-05
MsgTree: add MsgTreeElem.decode()

### LINKED ISSUE #660: Improve usability of message buffers as text
Since 1.8, message buffers are `bytes` end-to-end on Python 3. This is deliberate — remote command output is arbitrary and `MsgTreeElem.__str__` raises rather than decode implicitly — but it makes the most common library use case verbose, as noted in the review of PR #654:

```python
for buf, nodes in task.iter_buffers():
    print(NodeSet.fromlist(nodes), buf.message().decode())
```

The clush/clubak tools are not affected (`CLI/Display.py` decodes internally with `errors='replace'`); this concerns library users and documentation examples. With 1.11 requiring Python 3, two complementary options came out of the #654 discussion. They differ mostly in compatibility implications and time horizon.

### Option 1: a `decode()` method on `MsgTreeElem`

A thin proxy over `message().decode(...)`, mirroring `bytes.decode()`:

```python
for buf, nodes in task.iter_buffers():
    print(NodeSet.fromlist(nodes), buf.decode())
```

**Compatibility:** purely additive — no existing type or behavior changes, no application can break, and it is usable as soon as it ships; documentation examples can rely on it immediately. The one lasting commitment is the default `errors=` policy: once released it becomes an API contract, and changing it later would silently alter application behavior, so it has to be right the first time:

| Default | Behavior | Caveat |
|---|---|---|
| `'strict'` | mirrors `bytes.decode()` / `open()` / subprocess text mode | raises `UnicodeDecodeError` on any non-UTF-8 byte in command output |
| `'replace'` | never raises at decode or print time; same policy as clush/clubak display | lossy — invalid bytes become `U+FFFD` |
| `'surrogateescape'` | lossless: decode never raises and the original bytes round-trip via `.encode(enc, 'surrogateescape')` | lone surrogates are not encodable by a default `sys.stdout` (`errors='strict'`), so `print(buf.decode())` on non-UTF-8 output raises `UnicodeEncodeError` at print time |

In short: a display-oriented default favors `'replace'`; a processing-oriented default favors `'surrogateescape'`, with the print-time caveat above. Whatever the default, the alternatives should be documented in the docstring and Programming Guide.

### Option 2: opt-in text mode

A `text=True` / `encoding=` / `errors=` option mirroring `subprocess` and `open()`, and/or a task-level default encoding (e.g. `Task.set_default('encoding', 'utf-8')`), so that `ev_read` and `iter_buffers()` yield decoded `str`.

**Compatibility:** bytes stay the default, so applications that don't opt in see no change. The compatibility surface is in *how* the option is exposed:

- A **per-worker keyword argument** (appended at the end of signatures, per API convention) is explicit and local: only the code that passes `text=True` sees `str`. Safe, but needs to be threaded through every worker class.
- A **task-level default** is one line for the application, but acts at a distance: it changes the type delivered to *every* event handler in the task, including handlers inside third-party code built on ClusterShell that still expects `bytes`. A task mixing text and bytes workers also shares the task's message tree, raising a type-consistency question for `task.iter_buffers()` across workers.
- **Tree mode:** the gateway wire format stays bytes, with decoding on delivery at the head node — mixed-version trees (older gateways) are unaffected.
- The default `encoding`/`errors` values are the same lasting contract as in option 1.

This option needs a design pass before implementation (option surface, decode point — worker read path vs presentation time; per-line decoding is safe for UTF-8 since multibyte sequences cannot contain `\n`, but rules out UTF-16-family encodings).

### Time horizon

The options are not exclusive. Option 1 is available immediately and requires no migration; it remains valid and useful whether or not option 2 exists. Option 2 is a longer-term addition — it breaks nothing while opt-in, but going further and making text the *default* someday would be a breaking change for every application reading buffers, and could only be considered with a deprecation cycle in a future major release. In all cases `MsgTreeElem.__str__` keeps raising on Python 3.

thiell:
Implements option 1 of #660: a thin proxy over `message().decode()`, so the common library pattern becomes `buf.decode()` instead of `buf.message().decode()`.

- Defaults are `encoding='utf-8'`, `errors='replace'`: decoding and printing never raise, matching what clush/clubak already do internally (`CLI/Display.py`). The docstring documents `errors='surrogateescape'` as the lossless alternative, with the caveat that the result may not always be printable (lone surrogates vs a strict-errors stdout).
- Purely additive: no existing type or behavior changes, and `MsgTreeElem.__str__` still raises on Python 3.
- Library examples (README, Programming Guide, Task docstrings) updated to use it; the guide and docstring example patterns were re-run against a live ssh target.
- Also validated on Python 2.7 (not covered by CI): decode() returns `unicode` with the same defaults, keyword arguments work, and `errors='surrogateescape'` raises `LookupError` there — hence the "(Python 3)" note in the docstring.

Closes #660. Option 2 from the issue (opt-in text mode) is deliberately not included — it needs its own design pass and can get a dedicated issue if demand ever materializes.


degremont:
The one substantive issue: inconsistency with `node_buffer()` / `node_error()`
This is worth raising on the PR. The new `decode()` only lives on `MsgTreeElem`, but not every accessor returns one. `key_buffer`/`node_buffer` and `key_error`/`node_error` return plain bytes:


```python
def key_buffer(self, key):
    ...
    msgtree = self._msgtree('stdout')
    select_key = lambda k: k[1] == key
    return b''.join(bytes(msg) for msg in msgtree.messages(select_key))
node_buffer = key_buffer
```
So after this PR, `.decode()` behaves differently depending on which accessor you used:

* task.iter_buffers() → MsgTreeElem → .decode() uses the new default errors='replace' (undecodable bytes become U+FFFD).
* task.node_buffer(n) → bytes → .decode() is the builtin bytes.decode, i.e. errors='strict', which raises UnicodeDecodeError on the very same bytes.

The PR's own README even shows `task.node_buffer("green37").decode()`, which silently gets strict semantics — inconsistent with the `iter_buffers` example three lines below it. Same input, different behavior, no error, is exactly the kind of subtlety that bites users.

* CLI/Clush.py not migrated. It still calls `msg.decode()` directly (`self._outfiles[node].write("{}\n".format(msg.decode()))`). Not required for this PR, but note that switching it to msg.decode() on a MsgTreeElem would actually change clush's behavior from strict to replace — another instance of the consistency point above.


Not for later, an opt-in text mode could be useful:

opt-in "text mode" on the iterators (the `subprocess(text=True)` analog)
Let `iter_buffers()`/`iter_errors()` (and the worker-level and node variants) accept a text flag plus optional encoding/errors overrides. Default `text=False` preserves today's byte behavior exactly:

```python
def iter_buffers(self, match_keys=None, text=False, encoding=None, errors=None):
    for buf, keys in self._iter_msgtree('stdout', match_keys):
        if text:
            enc = encoding or self.default('encoding')
            err = errors or self.default('errors')
            yield buf.decode(enc, err), keys
        else:
            yield buf, keys      # unchanged: MsgTreeElem (bytes-like)
```
Usage becomes as clean as Python's:

```python
for out, nodes in task.iter_buffers(text=True):
    print(NodeSet.fromlist(nodes), out)     # out is already str
```

thiell:
I think I understand the problem with that method name indeed. How about renaming the method `text()` with the same implementation?
> `.text()` -> It's the ecosystem's word for exactly this: `pathlib.read_text()`, `requests.Response.text`, `subprocess(text=True)` — none of them named it decode, and for this reason.

The idea:
* 1.11 `MsgTreeElem.text(encoding='utf-8', errors='replace')`
* later: `text=True` (+`encoding/errors`) on the `Task`/`Worker` accessors and iterators, if demand shows up.

Let me know!

degremont:
Sounds like a better plan indeed

thiell:
Done and updated description.

degremont (approved):

PR #705, opened 2026-08-05
NodeUtils: export upcall variables to the environment

### LINKED ISSUE #704: groups.conf upcalls: ${GROUP:-default} expands from the caller's environment, not the group name
Group source upcall variables (`$GROUP`, `$NODE`, `$SOURCE`, `$CFGDIR`) are replaced textually with `string.Template.safe_substitute()`, which only recognizes the `$GROUP` and `${GROUP}` forms. Shell parameter expansions like `${GROUP:-default}` are left in the command unchanged, and the shell expands them from its own environment, where these variables are not set. With this generic `groups.conf`:

```ini
[Main]
default: test

[test]
map: echo "node-${GROUP:-default}"
```

the requested group name is silently ignored:

```console
$ cluset --groupsconf=bug.conf -f @test:foo
node-default        # expected: node-foo
```

and anything the caller exports under that name silently changes the result:

```console
$ GROUP=bar cluset --groupsconf=bug.conf -f @test:foo
node-bar            # expected: node-foo
```

In practice this form appears in map upcalls that query an external source by regex, using `${GROUP:-.*}` to default to match-all when the group name is empty — but which actually returns all nodes for any requested group. Same applies to `$NODE` in reverse upcalls.

Workaround: re-assign the substituted value to a shell variable first, e.g. `map: G="$GROUP"; mytool --match "${G:-.*}"`.

Reproduced with cluset/nodeset on master (1938944); present in all released versions. Related: #609.

thiell:
Upcall commands already inherit the caller's full environment. This change simply sets the four substitution variables (`$GROUP`, `$NODE`, `$SOURCE`, `$CFGDIR`) in that environment, so shell parameter expansions that are not replaced by the library, like `${GROUP:-default}`, expand to the same values instead of silently reading whatever the caller exported. `$GROUP` and `$NODE` are set empty when they do not apply to the upcall. Textual substitution is unchanged and still happens first, so existing configs behave identically.

This makes it easy for a map upcall to support a default when the group name is empty, e.g. for query-backed sources:

```ini
[prometheus]
map: promtool query -o json instant http://localhost:9090 "up{job=~\"${GROUP:-.*}\"}" | jq -r '.[].metric.instance' | cut -d: -f1
```

Here each Prometheus scrape job is a node group, resolving to the hostnames of its targets: `@prometheus:node_exporter` expands to the nodes scraped by the `node_exporter` job, and `@prometheus:` (empty group name) expands to the nodes of all jobs (useful for map-only sources, where `@prometheus:*` cannot work).

See #609 for a related earlier discussion (FYI).

Docs updated: groups.conf(5) + sphinx config.

Closes #704


thiell:
The docs now state it explicitly: exported variables take precedence over any same-named variables from the calling environment.

Flagging for the 1.11 release notes as a minor behavior change: upcall commands no longer inherit caller-exported `GROUP`, `NODE`, `SOURCE` or `CFGDIR` values.

lib/ClusterShell/NodeUtils.py
```diff
@@ -217,8 +217,11 @@ def _upcall_read(self, cmdtpl, args=dict()):
         """
         cmdline = Template(self.upcalls[cmdtpl]).safe_substitute(args)
```
degremont, line 218: if we start supporting variable interpolation through real shell variables (through environment), it seems we do not need the template anymore? The shell variable supports below will handle everything, no?

thiell: Nope. The shell doesn't expand variables inside single quotes, and single-quoted `$GROUP` is used in the field (just look at `man groups.conf`, single quotes are used in the example). So the environment export **complements** the template (for shell-side expansions like `${GROUP:-default}`) rather than replacing it.

degremont: OK, a bit weird, as now the expansion will happen sometimes under single quote, and sometimes not. User may be surprised.

thiell: `Template.safe_substitute()` is pure text replacement and has always ignored shell quoting. `$GROUP` is always replaced if found like that in the command text. The only surprise might come from existing `${GROUP:-default}` patterns **not** in single quote that were not resolved and that will then be resolved with this patch. It's actually the bug I am trying to fix here as I saw it in the field. :)

That said, I agree a single mechanism (shell-only) would probably be cleaner, but that breaks single-quoted `$GROUP` configs, including provided example configs and man pages. If we think it's a big deal (not sure), it's something we could consider to deprecate in a 2.0 maybe.

degremont: OK, so, what would be the less surprising to people over time would be for the variable to be 100% shell variables. So, let's keep the patch as-is, but, in the same time, we must change examples to use double quotes where appropriate. Let's write them as we were only supporting shell variables. We could deprecate that later.

thiell: Sounds good to me and actually that's a very good point: it would be possible to have an empty variable with this I think, so it is highly recommended to double-quote variables in upcalls. Best practice anyway ([SC2086](https://www.shellcheck.net/wiki/SC2086)). Thus, this will start a clean migration to shell-only variables in the future. I'll update this PR (docs, examples) with this.

doc/txt/groups.conf.txt
```diff
@@ -175,14 +177,14 @@ Simple configuration file for local groups and Slurm partitions binding.
 | autodir: /etc/clustershell/groups.d $CFGDIR/groups.d
 |
 | [local]
-| map: sed -n 's/^$GROUP:\\(.*\\)/\\1/p' /etc/clustershell/groups
+| map: sed -n "s/^$GROUP:\\(.*\\)/\\1/p" /etc/clustershell/groups
```
degremont, line 178: I was thinking of just changing this single quotes to double quotes actually :) not adding double quotes everywhere. That may be zealous.

tests/NodeSetGroupTest.py
```diff
@@ -1090,6 +1007,38 @@ def testConfigCFGDIR(self):
...
+        self.assertEqual(res.node_groups("node1"), ["g-node1"])
+        # exported variables take precedence over the caller's environment,
+        # and are exported empty to upcalls they do not apply to
+        os.environ['GROUP'] = 'callerenv'
+        try:
+            res = GroupResolverConfig(f.name)
+            self.assertEqual(str(NodeSet("@bar", resolver=res)), "n-bar")
+            self.assertEqual(res.all_nodes(), ["a-local"])
+        finally:
+            del os.environ['GROUP']
```
degremont: (minor) that deleted any pre-existing env variable with that name. Restore the actual state prior that test

caller could do:

GROUP=aaa
python test/NodeSetGroupTest.py....
echo $GROUP # empty

PR #706, opened 2026-08-05
RangeSetND: avoid quadratic padding scan in iter_padding()

thiell:
- `RangeSet.padding` is a property that walks the whole set; computing it for each yielded item made `iter_padding()` quadratic on large sets.
- Compute it once per vector instead; each item still gets its own list (contract now covered by a new test).
- Docstring now suggests `pads()` when per-item padding info is not needed.


tests/RangeSetNDTest.py
```diff
@@ -541,6 +541,20 @@ def test_pads(self):
...
+        rn0 = RangeSetND()
+        self.assertEqual([], list(rn0.iter_padding()))
+        rn1 = RangeSetND([['01-02', '003'], ['1-2', '4-5']])
+        self.assertEqual([(('1', '4'), [None, None]), (('1', '5'), [None, None]),
+                          (('2', '4'), [None, None]), (('2', '5'), [None, None]),
+                          (('01', '003'), [2, 3]), (('02', '003'), [2, 3])],
+                         list(rn1.iter_padding()))
+        # each item has its own padding list
+        items = list(rn1.iter_padding())
+        items[0][1][0] = 99
```
degremont: this line intent was not super clear.
```
assertIsNot(items[0][1], items[1][1])
```
may have been slightly clearer on the intent (insuring the pad list is not shared across multiple `yield`)

degremont (commented):
Could be nice to add some perf comparison in the commit message.

tests/RangeSetNDTest.py
thiell: Much clearer indeed, adopted thx!

thiell:
Added basic perf numbers to the commit message but here are the full results:

```
  ┌────────────┬────────┬─────────┬─────────┬─────────┐
  │   shape    │ items  │   old   │   new   │ speedup │
  ├────────────┼────────┼─────────┼─────────┼─────────┤
  │ 1D 1-100   │ 100    │ 536 µs  │ 15 µs   │ 35x     │
  ├────────────┼────────┼─────────┼─────────┼─────────┤
  │ 1D 1-1000  │ 1 000  │ 54.4 ms │ 151 µs  │ 360x    │
  ├────────────┼────────┼─────────┼─────────┼─────────┤
  │ 1D 1-10000 │ 10 000 │ 5.97 s  │ 1.5 ms  │ ~4000x  │
  ├────────────┼────────┼─────────┼─────────┼─────────┤
  │ 2D 10×10   │ 100    │ 108 µs  │ 12 µs   │ 9x      │
  ├────────────┼────────┼─────────┼─────────┼─────────┤
  │ 2D 32×32   │ 1 024  │ 3.45 ms │ 104 µs  │ 33x     │
  ├────────────┼────────┼─────────┼─────────┼─────────┤
  │ 2D 100×100 │ 10 000 │ 112 ms  │ 0.98 ms │ 114x    │
  └────────────┴────────┴─────────┴─────────┴─────────┘
```


from `bench_iterpad_2d.py`:
```py3
import sys, time
from ClusterShell.RangeSet import RangeSetND

def run_once(rs, reps):
    t0 = time.perf_counter()
    for _ in range(reps):
        for _item in rs.iter_padding():
            pass
    return time.perf_counter() - t0

for label, args, reps in (
        ('1D    100 items', [['1-100']], 1000),
        ('1D   1000 items', [['1-1000']], 50),
        ('1D  10000 items', [['1-10000']], 1),
        ('2D 10x10    100', [['1-10', '1-10']], 2000),
        ('2D 32x32   1024', [['1-32', '1-32']], 200),
        ('2D 100x100 10k ', [['1-100', '1-100']], 10),
):
    rs = RangeSetND(args)
    n = len(rs)
    best = min(run_once(rs, reps) for _ in range(3))
    print("%s  len=%-5d  %12.3f us/pass" % (label, n, best / reps * 1e6))
```

degremont (approved):

